PR #1, opened 2022-10-25
Moving Miden Examples to another repo

Dominik1999:


Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+std = ["hex/std", "miden/std", "rand-utils", "vm-core/std"]
+
+[dependencies]
+env_logger = { version = "0.9", default-features = false }
+hex = { version = "0.4", optional = true }
+log = { version = "0.4", default-features = false }
+miden = { package = "miden", version = "0.2.0", default-features = false }
+structopt = { version = "0.3", default-features = false }
+vm-core = { package = "miden-core", version = "0.2.0", default-features = false }
+rand-utils = { package = "winter-rand-utils", version = "0.2", optional = true }
```
bobbinth, line 33: minor nit: we usually try to have dependencies listed in alphabetical order.

src/comparison.rs
```diff
@@ -0,0 +1,62 @@
+use crate::Example;
+use log::debug;
+use assembly::Assembler;
+use miden::{ProgramInputs};
+
+// EXAMPLE BUILDER
+// ================================================================================================
+
+pub fn get_example(value: usize) -> Example {
```
bobbinth, line 9: We can probably get rid of this example as it is not super interesting.

Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "examples"
+version = "0.1.0"
+description = "Examples of using Miden VM"
+authors = ["miden contributors"]
+readme = "README.md"
+license = "MIT"
+repository = "https://github.com/maticnetwork/miden"
```
bobbinth, line 8: This should be updated to the current repo link.

src/conditional.rs
```diff
@@ -0,0 +1,64 @@
+use crate::Example;
+use log::debug;
+use assembly::Assembler;
+use miden::{ProgramInputs};
+
+// EXAMPLE BUILDER
+// ================================================================================================
+
+pub fn get_example(flag: usize) -> Example {
```
bobbinth, line 9: We can probably get rid of this example as it is not super interesting.

src/range.rs
```diff
@@ -0,0 +1,82 @@
+use crate::Example;
+use log::debug;
+use assembly::Assembler;
+use miden::{Program, ProgramInputs};
+use rand_utils::rand_vector;
+
+// EXAMPLE BUILDER
+// ================================================================================================
+
+pub fn get_example(num_values: usize) -> Example {
```
bobbinth, line 10: We can probably get rid of this example as it is not super interesting.

src/merkle.rs
```diff
@@ -0,0 +1,168 @@
+use crate::Example;
+use log::debug;
+use assembly::Assembler;
+use vm_core::{Felt, FieldElement, hasher, Program, ProgramInputs, StarkField};
+use rand_utils::prng_vector;
+
+// EXAMPLE BUILDER
+// ================================================================================================
+
+pub fn get_example(depth: usize) -> Example {
```
bobbinth, line 10: As is, I think this example is not super interesting. I wonder if we could transform it into a version of [Semaphore protocol](https://github.com/bobbinth/semaphore/blob/main/src/lib.rs). It could take a little bit of effort - but I think it would be very interesting how Semaphore implemented in Miden VM performs against native STARK implementation.

bobbinth (changes requested):
Looks good! I left a few small comments inline.

I think the bigger question is what we want the interface for this to look like. The current interface provides a CLI which can be used to run each example. While this works, it may make adding new examples a bit more challenging.

It would be cool to have examples in `.masm` files. But it would also be nice to have Rust code which checks the result of a computation. Not sure how to reconcile these two yet.

Dominik1999:
Closing this in favor of https://github.com/0xPolygonMiden/examples/pull/3

src/merkle.rs
tohrnii: @bobbinth The linked repo doesn't seem to be accessible. Is it private?

PR #2, opened 2022-10-27
Adding first examples

Dominik1999:
Adding old examples and the 4x4 Game of Life version as .masm and .inputs files to the repo. All are tested against the current next branch of Miden, which will soon be released as v0.3.0


bobbinth (approved):
Looks good! Thank you!

PR #3, opened 2022-10-27
Fix example builder to Miden next

Dominik1999:
Doesn't work yet, but all updated to Miden Next


src/collatz.rs
```diff
@@ -0,0 +1,82 @@
...
+        "
+    begin
+        push.0 adv_push.1 dup push.1 neq
+        while.true
+            # counter
+            swap push.1 add swap 
+            
+            # actual check
+            dup push.2 u32checked_mod
+            if.true
```
Dominik1999, line 28: @bobbinth I cannot push this PR forward, because also here I have the problem described in https://github.com/maticnetwork/miden/issues/466

If I remove the `if.true` procedure, it works like a charm. If I keep the procedure I get 

```
thread 'thread 'collatz::test_collatz_example_failcollatz::test_collatz_example' panicked at '' panicked at 'main transition constraint 1 did not evaluate to ZERO at step 28main transition constraint 1 did not evaluate to ZERO at step 28'
```

At lease `push.2 u32checked_mod` must always be either `0` or `1` so the `if.true` should not fail

PR #6, opened 2022-11-14
feat:adding a web assembly playground for the examples

Dominik1999:
Updated the Miden Assembly Playground from [here](https://github.com/timgestson/miden-assembly-playground) to accept inputs (`stack_init` and `advice_tape`) and to work with all our initial examples.


miden-wasm/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+
+
+/// We derive Deserialize/Serialize so we can persist app state on shutdown.
+#[derive(serde::Deserialize, serde::Serialize)]
+pub struct InputFile {
+    pub stack_init: Vec<String>,
+    pub advice_tape: Option<Vec<String>>,
+}
+
+fn get_program_inputs(stack_init: &[u64], advice_tape: &[u64]) -> ProgramInputs {
```
tohrnii, line 14: nit: It would be good to have doc comment for the method.

miden-wasm/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Parse stack_init vector of strings to a vector of u64
+fn stack_init(inputs_des: &InputFile) -> Vec<u64> {
+    inputs_des.stack_init
+        .iter()
+        .map(|v| v.parse::<u64>().unwrap())
+        .collect::<Vec<u64>>()
+}
+
+fn advice_tape(inputs_des: &InputFile) -> Vec<u64> {
```
tohrnii, line 26: nit: It would be good to have doc comment for the method.

miden-wasm/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub stack_init: Vec<String>,
+    pub advice_tape: Option<Vec<String>>,
+}
+
+fn get_program_inputs(stack_init: &[u64], advice_tape: &[u64]) -> ProgramInputs {
+    ProgramInputs::new(&stack_init, &advice_tape, Vec::new()).unwrap()
+}
+
+/// Parse stack_init vector of strings to a vector of u64
+fn stack_init(inputs_des: &InputFile) -> Vec<u64> {
```
tohrnii, line 19: minor nit (feel free to ignore): I think using `input` instead of `inputs_des` might be better.

miden-wasm/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+    //let options = ProofOptions::new(
+    //    32,
+    //    8,
+    //    0,
+    //    miden::HashFunction::Blake3_256,
+    //    miden::FieldExtension::None,
+    //    8,
+    //    256,
+    //);
+    //println!{"AdviceTape: {:?}", advice_tape};
```
tohrnii, line 63: Maybe these lines can be removed for now?

package.json
```diff
@@ -0,0 +1,55 @@
...
+  "version": "0.1.0",
+  "homepage": "https://github.com/0xPolygonMiden/examples",
+  "private": true,
+  "dependencies": {
+    "@codemirror/theme-one-dark": "^6.1.0",
+    "@mui/icons-material": "^5.6.1",
+    "@types/react": "^18.0.4",
+    "@types/react-dom": "^18.0.0",
+    "@uiw/react-codemirror": "^4.6.0",
+    "list-react-files": "^0.2.0",
```
tohrnii, line 12: Is this import required?

.gitignore
```diff
@@ -1,10 +1,23 @@
...
-/target/
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
-# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
-# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
-Cargo.lock
+# dependencies
+/node_modules
+/.pnp
+.pnp.js
```
tohrnii, line 6: Are these required?

miden-wasm/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+        .collect::<Vec<u64>>()
+}
+
+
+#[wasm_bindgen]
+pub fn program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64> {
+    console_error_panic_hook::set_once();
+    wasm_logger::init(wasm_logger::Config::default());
+    log::info!("adad");
+
```
tohrnii, line 41: minor nit: Can this line be removed?

src/App.tsx
```diff
@@ -0,0 +1,194 @@
...
+        value.push(options[i].value);
+      }
+    }
+    setExample(value)
+    
+    const inputs = await getExample(value, true)
+    setInputs(inputs)
+    
+    const code = await getExample(value, false)
+    setCode(code)
```
tohrnii, line 84: We could return an object with `inputs` and `code` instead of calling the function twice.

src/App.tsx
```diff
@@ -0,0 +1,194 @@
...
+async function getExample(example: string[], inputs: Boolean) {
+  if (inputs) {
+    const response = fetch(`https://raw.githubusercontent.com/0xPolygonMiden/examples/main/examples/${example}.inputs`)
+    return (await response).text();
+  }
+  else {
+    const response = fetch(`https://raw.githubusercontent.com/0xPolygonMiden/examples/main/examples/${example}.masm`)
+    return (await response).text();
+  }
+}
```
tohrnii, line 31: We should probably return an object with inputs and code instead of calling the function twice and taking inputs boolean flag.

src/App.tsx
```diff
@@ -0,0 +1,194 @@
+import {
+  AppBar,
+  Alert,
+  Box,
+  Button,
+  Container,
+  FormGroup,
```
tohrnii, line 7: nit: The linter complains that this import is unused so maybe we should remove it.

tohrnii (changes requested):
Thank you @Dominik1999, looks good! I left a few minor comments inline. This is not a full review yet. I still need to figure out a few things.

Also, we should probably not use both `yarn` and `npm`. We should stick to one and remove the other lock file. I don't understand the nuances well enough to have an opinion so we can go with whichever of the two you think is best.

src/App.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const [numOfOutputs, setNumOfOutputs] = React.useState(1);
+
+  // this should be dyncamically read from ./examples/ - buy my TS knowledge is limited
+  const examples = [
+    'collatz',
+    'comparison',
+    'conditional',
+    'fibonacci',
+    'game-of-life-4x4',
+    'nPrime',
```
tohrnii, line 65: The nPrime example seems to be missing from the examples directory. We should either add it in the examples or remove the entry from here.

README.md
```diff
@@ -1,2 +1,56 @@
...
+
+## Complex examples - larger complex nested operations
+
+- **Game-of-Life**: Implementation of Conway's Game of Life. But we can prove it. The static example runs on a 4x4 cell universe and 1000 generations.
+
+
+
+---
+
+We use a fork of https://github.com/timgestson/miden-assembly-playground
```
bobbinth, line 33: I would put this into a separate section towards the end - maybe "Acknowledgement" or something like this?

README.md
```diff
@@ -1,2 +1,56 @@
...
+
+[Use the playground!](https://0xpolygonmiden.github.io/examples/)
+
+The goal of this playground is for developers to see how easy it is to write and execute code in Miden Assembly. The examples come from the community and the team. If you come up with an example of your own, we are happy to include it here to the list.
+
+Examples are written in Miden Assembly, see [Miden Assembly Docu](https://maticnetwork.github.io/miden/user_docs/assembly/main.html). Miden Assembly programs can have two different inputs. They can have a `stack_init` as public input and the `advice_tape` as secret input. 
+
+Next to running examples you can also write your own program and execute it. The examples can then serve as inspiration. 
+
+## Simple examples - to see how the code works
```
bobbinth, line 13: Maybe put this under a common section - something like "Available examples"?

README.md
```diff
@@ -1,2 +1,56 @@
...
+
+## Prerequisites
+
+Install [wasm-pack](https://rustwasm.github.io/wasm-pack/installer/)
+
+## Starting site
+
+We recomment using the link above to play with the examples. 
+
+If you want to run it locally, then (in the project directory) run:
```
bobbinth, line 43: This feels like it should be a separate section which describes how to run the site locally - so, maybe put this under "Running the playground locally" section?

bobbinth (commented):
I did a quick read (mainly through the readme file) and left a couple of small comments inline.

Also, we should probably add a License section at the end (similar to how we do it in the main Miden repo).

Another comment for the future: I would probably prefer using [Deno](https://github.com/denoland/deno) over Node.js. I believe (though could be wrong), it would make the repo cleaner and Deno also has very good integration with Rust.

.gitignore
```diff
@@ -1,10 +1,23 @@
-# Generated by Cargo
-# will have compiled files and executables
-/target/
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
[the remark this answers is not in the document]
Dominik1999: done

README.md
```diff
@@ -1,2 +1,56 @@
...
-Examples of using Polygon Miden
+# Miden Assembly Playground
+
+Playground for example programs for [Miden](https://github.com/maticnetwork/miden) in Miden Assembly.
+
+[Use the playground!](https://0xpolygonmiden.github.io/examples/)
+
+The goal of this playground is for developers to see how easy it is to write and execute code in Miden Assembly. The examples come from the community and the team. If you come up with an example of your own, we are happy to include it here to the list.
+
+Examples are written in Miden Assembly, see [Miden Assembly Docu](https://maticnetwork.github.io/miden/user_docs/assembly/main.html). Miden Assembly programs can have two different inputs. They can have a `stack_init` as public input and the `advice_tape` as secret input.
```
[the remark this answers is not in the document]
Dominik1999: done

miden-wasm/src/lib.rs
Dominik1999: What is a doc comment?

miden-wasm/src/lib.rs
Dominik1999: what is a doc comment?

miden-wasm/src/lib.rs
Dominik1999: yes indeed. Would be cool to be able to prove later though

miden-wasm/src/lib.rs
Dominik1999: indeed. Was a leftover

src/App.tsx
Dominik1999: Hmm, but we use all of them but `FormGroup`. I removed that now

tohrnii: @Dominik1999 Ah sorry, I should have clarified. Github automatically marks additional 3 lines above (probably to make locating the code easier). I did mean only `FormGroup`.

miden-wasm/src/lib.rs
grjte: https://doc.rust-lang.org/rust-by-example/meta/doc.html

src/App.tsx
```diff
@@ -0,0 +1,187 @@
...
+                }}
+              >
+                {examples.map((name) => (
+                  <option key={name} value={name}>
+                    {name}
+                  </option>
+                ))}
+              </Select>
+          </FormControl>
+        </Box>
```
grjte, line 113: We can improve this `Select` by following the MaterialUI `Select` component pattern. I also think we should move the outputs and execute button to the top and make a few minor adjustments to the styling. I propose:

1. Delete everything below the source code `CodeMirror`
2. Replace this highlighted box with the box code below.
3. Replace `handleChangeMultiple` with `handleSelectChange` code below.
4. Change the type of `example` and the input to `getExample` to `string` instead of `string[]`

```
<Box sx={{ display: 'flex', flexDirection: 'row', mx: 'auto', my: 4}}>
          <FormControl variant="outlined" sx={{minWidth: 120}}>
              <InputLabel shrink htmlFor="select-example">
                Example
              </InputLabel>
              <Select
                value={example}
                onChange={handleSelectChange}
                labelId='select-example'
              >
                {examples.map((name) => (
                  <MenuItem value={name} key={name}>{name}</MenuItem>
                ))}
              </Select>
          </FormControl>
          <FormControl variant="outlined" sx={{minWidth: 120, mx: 2}}>
              <InputLabel shrink htmlFor="select-outputs">
                Outputs
              </InputLabel>
              <Select
              value={numOfOutputs}
              onChange={(e) => setNumOfOutputs(Number(e.target.value))}
              labelId='select-outputs'
            >
              <MenuItem value={0} key={0}>0</MenuItem>
              <MenuItem value={1} key={1}>1</MenuItem>
              <MenuItem value={16} key={16}>16</MenuItem>
            </Select>
          </FormControl>
          <Button
            variant="contained"
            color="primary"
            onClick={() => {
              init().then(() => {
                try {
                  let resp = program(code, inputs, numOfOutputs);
                  console.log(resp);
                  setOutput({
                    init: true,
                    success: true,
                    text: resp.join(" "),
                  });
                } catch (error) {
                  setOutput({
                    init: true,
                    success: false,
                    text: "Error: Check the developer console for details.",
                  });
                }
              });
            }}
            endIcon={<SendIcon />}
          >
            Execute
          </Button>
          {output.init ? (
            <Alert
              variant="outlined"
              severity={output.success ? "success" : "error"}
              sx={{mx: 2}}
            >
              <Typography color="inherit" component="div">
                {output.text}
              </Typography>
            </Alert>
          ) : (
            <></>
          )}
        </Box>
```

```
  const handleSelectChange = async (event: SelectChangeEvent) => {
    const value = event.target.value;
    // set the current example to the selected one
    setExample(value)

    // retrieve the example data and update the inputs and code
    const example_code = await getExample(value)
    setInputs(await example_code[0])
    setCode(await example_code[1])    

    // reset the output
    setOutput(emptyOutput)
  }
```

src/App.tsx
```diff
@@ -0,0 +1,187 @@
...
+  add
+end`
+
+  const [code, setCode] = React.useState(exampleCode);
+
+  const [output, setOutput] = React.useState({
+    init: false,
+    success: true,
+    text: "",
+  });
```
grjte, line 48: let's add a constant for empty output here so we can reset it when the example is changed:

```

  const emptyOutput = {
    init: false,
    success: true,
    text: "",
  };
  const [output, setOutput] = React.useState(emptyOutput);
```

grjte (changes requested):
Thanks @Dominik1999, this is really cool!

I left comments inline about, including some proposed changes to tidy up the ui/ux. Let me know if you have any questions!

Also, not all of the examples are working for me. I'm getting a few 404s. Can you check on this?

src/App.tsx
grjte: tbh, it actually probably makes sense to define this as its own component and then put it at both the top and the bottom for better UX on the long examples

README.md
```diff
@@ -1,2 +1,63 @@
...
-Examples of using Polygon Miden
+# Miden Assembly Playground
+
+Playground for example programs for [Miden](https://github.com/maticnetwork/miden) in Miden assembly.
+
+[Use the playground!](https://0xpolygonmiden.github.io/examples/)
+
+The goal of this playground is for developers to see how easy it is to write and execute code in Miden assembly. The examples come from the community and the team. If you come up with an example of your own, we are happy to include it here to the list.
+
+Examples are written in Miden assembly, see [Miden assembly Documentation](https://maticnetwork.github.io/miden/user_docs/assembly/main.html). Miden assembly programs can have two different inputs. They can have a `stack_init` as public input and the `advice_tape` as secret input.
```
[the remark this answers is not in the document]
Dominik1999: Got it, one question, though. [Here](https://maticnetwork.github.io/miden/intro/overview.html#inputs-and-outputs) we say we can have two type of external inputs and I think the advice provider can entail three different kinds of secret inputs, right?

README.md
```diff
@@ -1,2 +1,63 @@
...
+
+## Starting site
+
+We recomment using the link above to play with the examples. 
+
+If you want to run it locally, then (in the project directory) run:
+
+`npm install && npm start`
+
+Runs the app in the development mode.\
```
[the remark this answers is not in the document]
Dominik1999: good point

miden-wasm/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "miden-wasm"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+console_error_panic_hook = { console_error_panic_hook = "0.1.7" }
+miden = { git = "https://github.com/maticnetwork/miden", branch = "next", package = "miden", default-features = false}
```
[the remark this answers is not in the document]
Dominik1999: I think it makes most sense to have the playground on the latest stable to be honest

src/App.tsx
```diff
@@ -0,0 +1,187 @@
...
+  Box,
+  Button,
+  Container,
+  FormControl,
+  InputLabel,
+  MenuItem,
+  Select,
+  Toolbar,
+  Typography,
+} from "@mui/material";
```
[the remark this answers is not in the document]
Dominik1999: I will open an issue that this will be the first refactoring

src/App.tsx
```diff
@@ -0,0 +1,187 @@
...
+import SendIcon from "@mui/icons-material/Send";
```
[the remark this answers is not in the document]
Dominik1999: I will add this to the refactoring issue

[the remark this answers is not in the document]
Dominik1999: https://github.com/0xPolygonMiden/examples/issues/7

src/App.tsx
```diff
@@ -0,0 +1,187 @@
...
+import React from "react";
+import CodeMirror from "@uiw/react-codemirror";
+import { oneDark } from "@codemirror/theme-one-dark";
+import init, { program } from "miden-wasm";
+import "./App.css";
```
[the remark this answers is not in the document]
Dominik1999: Jep we can delete that

src/theme.ts
```diff
@@ -0,0 +1,19 @@
...
+      main: "#0052cc",
+    },
+    secondary: {
+      main: "#edf2ff",
+    },
+    background: {
+      default: "#f5f5f5",
+    },
+  },
+});
```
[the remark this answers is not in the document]
Dominik1999: How nice of you that you already gave me the right color codes here!

Dominik1999:
Thank so much for your review @grjte. The two non-working examples should work when we rabase and merge.

Dominik1999 (commented):
Thanks for the review!

README.md
```diff
@@ -1,2 +1,55 @@
...
+
+- **Fibonnacci**: Elegant way to calculate the 1000th fibonacci number. 
+
+## Complex examples - larger complex nested operations
+
+- **Game-of-Life**: Implementation of Conway's Game of Life. But we can prove it. The static example runs on a 4x4 cell universe and 1000 generations.
+
+---
+# Running the playground locally
+# Running the playground locally
```
bobbinth, line 35: The title is repeated here. Also, I would use first level title (i.e., `#`) only at the very top (for "Miden Assembly Playground") and would "downgrade" all other titles. For example, here I'd use second-level title (i.e., `##`).

bobbinth:
One other comment: I wonder whether rather than putting the playground into the root of the repo it should go into something like `/wasm-playground` (or just `/playground`) directory. This would enable us to add other things in the future (e.g., CLI mentioned in #9). Maybe the structure of the repo could be something like:

```
├─ bench-cli
├─ examples
└─ playground
```

What do you guys think?

Dominik1999:
Thanks @bobbinth ! 

I think it makes sense. Maybe we can merge this PR as is and I open a issue to change the structure and add the CLI?

src/App.tsx
```diff
@@ -0,0 +1,184 @@
...
+            height="100%"
+            theme={oneDark}
+            onChange={setInputs}
+          />
+        </Box>
+        <Box sx={{ my: 4 }}>
+        <Typography color="primary" variant="h6">Miden Assembly Code</Typography>
+          <CodeMirror
+            value={code}
+            height="100%"
```
grjte, line 173: As suggested by @tohrnii, let's set a max height here. I think 500px is ok. We'll revisit later when upgrading the design anyway

```
maxHeight="500px"
```

src/App.tsx
```diff
@@ -0,0 +1,184 @@
...
+            </Alert>
+          ) : (
+            <></>
+          )}
+        </Box>
+        <Box sx={{ my: 4 }}>
+        <Typography color="primary" variant="h6">Inputs</Typography>
+          <CodeMirror
+            value={inputs}
+            height="100%"
```
grjte, line 164: Let's set a max height here. I think 200px is probably ok for this one.

src/App.tsx
```diff
@@ -0,0 +1,184 @@
...
+              sx={{mx: 2}}
+            >
+              <Typography color="inherit" component="div">
+                {output.text}
+              </Typography>
+            </Alert>
+          ) : (
+            <></>
+          )}
+        </Box>
```
grjte, line 159: For the future: It would be good to pull this out into its own component and add it at both the top and the bottom of the screen. I think that would provide better UX (people wouldn't need to scroll back up after writing their code). Do you agree? If so, can you create an issue?

grjte (changes requested):
Thanks @Dominik1999! I have a few more small things that could use adjustment, as well as a failure when I run `wasm-pack test --node`

grjte:
> One other comment: I wonder whether rather than putting the playground into the root of the repo it should go into something like `/wasm-playground` (or just `/playground`) directory.

I agree with the structure suggestion too, but I think it's ok to merge this once the small points are addressed so there's something to show for the discussion today. Please update here with the issue link for restructure + CLI once you've created it @Dominik1999

src/App.tsx
Dominik1999: good point, somehow that slipped

src/App.tsx
Dominik1999: Yes, I will create an issue

Dominik1999:
Thanks @grjte 

here is the issue for changing the repo structure: https://github.com/0xPolygonMiden/examples/issues/11

README.md
```diff
@@ -1,2 +1,58 @@
...
+### Math examples - see more complex numerical calculations 
+
+- **Collatz**: A program which executes an unbounded loop to compute a Collatz sequence which starts with the provided value; the output of the program is the number of steps needed to reach 1 - the end of the sequence.
+
+- **Fibonnacci**: Elegant way to calculate the 1000th fibonacci number. 
+
+### Complex examples - larger complex nested operations
+
+- **Game-of-Life**: Implementation of Conway's Game of Life. But we can prove it. The static example runs on a 4x4 cell universe and 1000 generations.
+
```
grjte, line 34: nit: `nPrime` is missing from here. We could add something like:

"nPrime: A program which calculates the nth prime number, using the VM memory."

grjte (approved):
Looks good, thank you! I left one final comment that would be good to update before merging, but I'm approving so you're not blocked.

Can you please make sure that everything we discussed creating issues for gets created? (e.g. there's a TODO in the code that doesn't seem to be tracked by an issue yet, unless I just missed the issue when I looked?)

README.md
Dominik1999: it is there now, thanks again!

grjte (approved):

tohrnii (approved):
All looks good! Thank you!

Issue #11, opened 2022-11-25
Change repo structure and put playground into own folder


Dominik1999:
One other comment: I wonder whether rather than putting the playground into the root of the repo it should go into something like `/wasm-playground` (or just `/playground`) directory. This would enable us to add other things in the future (e.g., CLI mentioned in #9). Maybe the structure of the repo could be something like:

```
├─ bench-cli
├─ examples
└─ playground
```

What do you guys think?

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/examples/issues/6#issuecomment-1326857075_

grjte:
Once we do this, let's also see about deploying the playground to https://0xpolygonmiden.github.io/playground instead of to `examples`

bobbinth:
Should we try to put this under a custom domain? Something like `miden.polygon.technology/vm-playground`

grjte:
Yes, we could change the domain in the settings easily, and I think `miden.polygon.technology` would be good. The path would still depend on these changes though. I prefer `playground` to `vm-playground`

Also, I would prefer to complete this issue before releasing v0.1. What do you think?

PR #20, opened 2022-11-25
feat: adding GH workflow file

Dominik1999:
Adding a workflow for GH pages to show the examples playground


.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,31 @@
+name: Deploy Miden Assembly playground
+
+on:
+  push:
+    branches:
+      - test-playground-deployment
```
grjte, line 6: This should be either `main` or `next`, depending on when we want to deploy. It looks like we don't have a `next` branch yet, so maybe set it to `main`

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "miden-assembly-examples-playground",
   "version": "0.1.0",
-  "homepage": "https://github.com/0xPolygonMiden/examples",
+  "homepage": "https://0xpolygonmiden.github.io/examples/",
```
grjte, line 4: Maybe we should name this `/playground` instead of `/examples`? What do you think?

webpack.config.js
```diff
@@ -1,18 +1,39 @@
...
+      {
+        test: /\.css$/,
         exclude: /node_modules/,
+        use:['style-loader', 'css-loader'],
       },
     ],
   },
   resolve: {
-    extensions: ['.tsx', '.ts', '.js'],
+    extensions: ['*', '.tsx', '.ts', '.js'],
```
grjte, line 36: Do we need all extensions? I'd prefer to specify the ones we need, rather than using `*`

grjte (changes requested):
Thanks @Dominik1999, looks good! I left a few small things inline.

webpack.config.js
Dominik1999: Unfortunately, this is like black magic to me, and after 30 failed attempts, this config worked. Let me try.

package.json
Dominik1999: The GitHub Pages repository URL is always https://{userid}.github.io/{reponame} The GitHub Pages URL is based on the GitHub repo URL, and takes the format https://{userid}.github.io/{reponame} . The moment you use it, GitHub Pages will generate a website based on the contents of your repo.

Let's keep it like is for now.

webpack.config.js
Dominik1999: I changed and try to build it

grjte: Looks good, thanks!

package.json
grjte: I see - yes, I thought it would be more straightforward. I think we can do this from a subdirectory, so once we change the directory structure in #11 we should be able o deploy to `/playground`

grjte (changes requested):
Hey, the changes look good. However, there's a typo in `fibonacci.masm`. Can you correct the file name? Then we can merge.

grjte (changes requested):
There's still a typo in `fibonacci.masm`

Dominik1999:
Now it should be good. Sorry to many things in parallel it seems

grjte (approved):
Looks good now, thanks @Dominik1999!

PR #22, opened 2022-11-29
fix: nPrime to nprime to avoid 404

Dominik1999:
typo in nprime example list


tohrnii (approved):
Looks good! Thank you!

PR #23, opened 2022-11-29
fix: final output of GoL was reverse

Dominik1999:
The stack output of game of life is reversed. This PR fixes this


grjte (approved):
Looks good!

Issue #24, opened 2022-11-29
nprime example returns error


Dominik1999:
When executing the nprime example in the playground, I get

```
panicked at 'called `Result::unwrap()` on an `Err` value: "Failed to generate exection trace = DivideByZero(173), and advice_tape = []"', src/lib.rs:60:17
```

PR #26, opened 2022-11-30
fix: nprime example remove drop

Dominik1999:
The nprime example was broken, and this is a fix. Fixes https://github.com/0xPolygonMiden/examples/issues/24

It was already fixed on Miden main 8 days ago in here https://github.com/maticnetwork/miden/commit/7dd466c2649c21f7eb2ce96f1206ff753272ac5a

Somehow we should avoid maintaining this file in two places.

### DIFF
--- a/examples/nprime.masm
+++ b/examples/nprime.masm
@@ -11,7 +11,6 @@ proc.append
 
     # [prime, i, n, primes..]
     mem_store
-    drop
 
     # [i++, n, primes..]
     swap.2


grjte (approved):
Approved! We've removed all but 2 examples from the main vm repo, so there's minimal repetition. However, this one and fibonacci are still there. Maybe it's worth removing this one as well and updating the readme & docs in the vm to link to this examples repo and the playground? If you agree you could open an issue for discussion in the main miden vm repo and see what others think.

Issue #25, opened 2022-11-29
Game of Life Example doesn't work as expected


Dominik1999:
When I run the Game of Life example in the playground, I would expect using the initial universe

```
"stack_init": ["0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "1", "1", "0", "0", "1", "0"]
```

the following output

```
["0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "1", "1", "0", "0", "1", "1"]
```

But somehow I get only 0's.

PR #27, opened 2022-11-30
fix: remove drop from game of life example

Dominik1999:
Fixes https://github.com/0xPolygonMiden/examples/issues/25

The problem was that it was recently introduced that `loc.store.i` also drops an element from stack. 

Also the order of the output stack is already reversed, so now it is correct.

### DIFF
--- a/examples/game-of-life-4x4.masm
+++ b/examples/game-of-life-4x4.masm
@@ -7,42 +7,42 @@
 
 # We store the initial configuration
 proc.storecellsn.16
-    loc_store.0 drop
-    loc_store.1 drop
-    loc_store.2 drop
-    loc_store.3 drop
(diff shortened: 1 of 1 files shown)


tohrnii (commented):
There are a lot of instances of `loc_store` with a subsequent `drop` instruction as well. Can you please check that they don't need to be removed?
Also, if I understand this correctly this is run for 1000 generations by default currently right? Should we run it for a lower number of generations? Maybe 10?

tohrnii:
It seems to me like the other subsequent drops don't result in incorrect stack, they are just redundant.

Dominik1999:
very nice! Thanks. Then I will remove them and set the default to 10

PR #32, opened 2022-12-16
docs: add CHANGELOG.md

Dominik1999:


CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
+
+## 0.1.0 (2021-12-16)
+
+- Initial release of the WASM Miden playground
+- Adding six examples for written in Miden assembly
+- Adding a simple front to show the Miden assembly code and execute it
```
grjte, line 7: 1. I think there's either an extra word (for) or a missing word in the second point
2. I would say "front-end" instead of "front"
3. nit: I would probably use past tense (e.g. "Added ..."), since that's what we do in the Miden VM and Winterfell changelogs

CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
+
+## 0.1.0 (2021-12-16)
```
grjte, line 3: There's a typo on the year

grjte (changes requested):
Thanks @Dominik1999! I left a few minor points inline

CHANGELOG.md
Dominik1999: haha thanks a lot and sorry, I could have read it again before submitting...

grjte (approved):
Looks good, thank you!

Issue #9, opened 2022-11-24
Create a small benchmarking CLI


bobbinth:
We should add a small CLI which could be used to benchmark any of the examples. This would also be helpful for the Delendum benchmarking project.

The CLI should take a path to the example directory as input, execute/prove the example, and output the following metrics:

1. Cold compilation time (i.e., when the example is compiled with a new assembler).
2. Hot compilation time (i.e., when the example is complied with an assembler which was already used to compile the same code).
3. Execution time.
4. Proving time.
5. Verification time.

It would also be nice to enable benchmarking at 96-bit and 128-bit security levels. So, maybe this could be another input parameter for the CLI.

Lastly, it would be really nice to also have the benchmark verify the correctness of the result. The simplest way to do it would be include the expected result for a given example into the example directory. This would, however, mean that adjusting example parameters would be difficult. We could also include a small Rust file with each example which could output the expected result for a given set of parameters - but this might be something we do later on.

Dominik1999:
We have a script for benchmarking here in the Delendum Project folder. https://github.com/delendum-xyz/zk-benchmarking

The question is, do we need our own script and the ability to benchmark in our examples-repo as well?

bobbinth:
I think so. I took a quick look at the benchmarking program, and I don't think it has everything we need (but maybe I missed it)?

But also, it would be good to have a separate benchmark CLI which we could evolve independently in the future.

Dominik1999:
Ok, I am on this issue now. 

Is this related to the Delendum benchmarking examples? Or do you mean the examples from the Playground?

bobbinth:
I think this is more general - i.e., would be good to be able to benchmark all examples in this repo.

Dominik1999:
> Lastly, it would be really nice to also have the benchmark verify the correctness of the result. The simplest way to do it would be include the expected result for a given example into the example directory. This would, however, mean that adjusting example parameters would be difficult. We could also include a small Rust file with each example which could output the expected result for a given set of parameters - but this might be something we do later on.

I think we had this before in the examples. I will create another issue to track that part.

PR #33, opened 2022-12-19
feat: adding a simple benchmarking CLI for the examples

Dominik1999:
Closes https://github.com/0xPolygonMiden/examples/issues/9

### DIFF
--- a/benchmarking-cli/.gitignore
+++ b/benchmarking-cli/.gitignore
@@ -1,0 +1,3 @@
+pkg/
+target/
+Cargo.lock
--- a/benchmarking-cli/README.md
+++ b/benchmarking-cli/README.md
@@ -1,0 +1,37 @@
+# Benchmarking CLI for Miden Examples
+
+This is a simple CLI that enables users to benchmark the Miden examples. You can run any example in https://github.com/0xPolygonMiden/examples/tree/main/examples. 
+
+## Usage
+To run the fibonacci example, you can either run
+
+```
+cargo build
+```
+
+and then 
+
+```
+./target/debug/miden-benchmarking-cli --example fibonacci
+```
+
+OR you run 
+
+```
+cargo run -- -example fibonacci   
+```
+
+You can pass two additional parameters to the CLI `security` and `output`. `security` can be `"high"` for 128-bit security and will default to 96-bit. `output` defines the number of stack outputs the program returns. It defaults to 1. 
+
+In general the CLI works as follows:
+
+`miden-benchmarking-cli --example <EXAMPLE> --security <SECURITY>`
+
+```
+Options:
+  -e, --example <EXAMPLE>    Provide example name as in ../examples
+  -s, --security <SECURITY>  Set to 'high' if 128-bit is needed [default: ]
+  -o, --output <OUTPUT>      Set the number of desired stack outputs [default: 1]
+  -h, --help                 Print help information
+  -V, --version              Print version information
+```
--- a/benchmarking-cli/src/main.rs
+++ b/benchmarking-cli/src/main.rs
@@ -1,0 +1,159 @@
+use clap::Parser;
+use miden::{Assembler, ProgramInputs, ProofOptions};
+use winter_math::StarkField;
+use std::fs;
+use std::time::Instant;
+
+#[derive(serde::Deserialize, serde::Serialize)]
+pub struct InputFile {
+    pub stack_init: Vec<String>,
+    pub advice_tape: Option<Vec<String>>,
+}
+
+// Creates a `ProgramInputs` struct for Miden VM from the specified inputs.
+// ToDo: we should allow other types of advice inputs as well. We'll want to make the same update to the playground as well.
+fn get_program_inputs(stack_init: &[u64], advice_tape: &[u64]) -> ProgramInputs {
+    ProgramInputs::new(stack_init, advice_tape, Vec::new()).unwrap()
+}
+
+// Parse stack_init vector of strings to a vector of u64
+fn get_stack_init(inputs: &InputFile) -> Vec<u64> {
+    inputs
+        .stack_init
+        .iter()
+        .map(|v| v.parse::<u64>().unwrap())
+        .collect::<Vec<u64>>()
+}
+
+// Parse advice_tape vector of strings to a vector of u64
+fn get_advice_tape(inputs: &InputFile) -> Vec<u64> {
+    inputs
+        .advice_tape
+        .as_ref()
+        .unwrap_or(&vec![])
+        .iter()
+        .map(|v| v.parse::<u64>().unwrap())
+        .collect::<Vec<u64>>()
+}
+
+#[derive(Parser)]
+#[clap(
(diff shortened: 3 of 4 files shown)


benchmarking-cli/.gitignore
```diff
@@ -0,0 +1,3 @@
+pkg/
+target/
+Cargo.lock
```
itzmeanjan, line 3: Missing newline at end of file.

itzmeanjan (approved):
Looks good.

benchmarking-cli/.gitignore
Dominik1999: thanks

benchmarking-cli/README.md
```diff
@@ -0,0 +1,37 @@
...
+./target/debug/miden-benchmarking-cli --example fibonacci
+```
+
+OR you run 
+
+```
+cargo run -- -example fibonacci   
+```
+
+You can pass two additional parameters to the CLI `security` and `output`. `security` can be `"high"` for 128-bit security and will will default to 96-bit. `output` defines the number of stack outputs the program returns. It defaults to 1.
```
itzmeanjan, line 24: ~will will default to~ -> will default to

benchmarking-cli/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+        .unwrap();
+
+    println! {"Execution Time: {} ms", now.elapsed().as_millis()}
+
+    // Proving time
+    let mut proof_options = ProofOptions::with_96_bit_security();
+
+    if args.security == "high" {
+        proof_options = ProofOptions::with_128_bit_security();
+    }
```
itzmeanjan, line 115: You can probably do

```rust
let proof_options = if args.security == "high" { ProofOptions::with_128_bit_security() } else { ProofOptions::with_96_bit_security() };
```

itzmeanjan (approved):
Looks good !

itzmeanjan:
@Dominik1999 Issuing `cargo fmt` should resolve issues related to source formatting.

benchmarking-cli/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+#[derive(serde::Deserialize, serde::Serialize)]
+pub struct InputFile {
+    pub stack_init: Vec<String>,
+    pub advice_tape: Option<Vec<String>>,
+}
+
+// Creates a `ProgramInputs` struct for Miden VM from the specified inputs.
+fn get_program_inputs(stack_init: &[u64], advice_tape: &[u64]) -> ProgramInputs {
+    ProgramInputs::new(stack_init, advice_tape, Vec::new()).unwrap()
+}
```
grjte, line 15: Comment for the future: we should allow other types of advice inputs as well. We'll want to make the same update to the playground as well.

benchmarking-cli/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+        default_value("1")
+    )]
+    output: usize,
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // The CLI should take a path to the example directory as input, execute/prove the example, and output the following metrics:
+
+    // Cold compilation time (i.e., when the example is compiled with a new assembler).
+    // Hot compilation time (i.e., when the example is complied with an assembler which was already used to compile the same code).
```
grjte, line 74: nit: this should be "compiled" instead of "complied"

benchmarking-cli/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+    // Verification time
+    let program_input_u64 = inputs
+        .stack_init()
+        .iter()
+        .map(|x| x.inner())
+        .collect::<Vec<u64>>();
+
+    let now = Instant::now();
+    let _result = miden::verify(program.hash(), &program_input_u64, &output, proof)
+        .map_err(|err| format!("Program failed verification! - {}", err));
```
grjte, line 150: If we're not using the result, we should be able to get rid of the `let _result =`, right?

benchmarking-cli/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    // let's read the program
+    let program_string = fs::read_to_string(format!("../examples/{}.masm", &args.example))?;
+
+    // let's read the input fils
+    let input_string = fs::read_to_string(format!("../examples/{}.inputs", &args.example))
+        .map_err(|err| format!("Failed to open input file `{}` - {}", &args.example, err))?;
+    let inputs_des: InputFile = serde_json::from_str(&input_string)
+        .map_err(|err| format!("Failed to deserialize input data - {}", err))
+        .unwrap();
```
grjte, line 93: I'd prefer `expect` instead of `unwrap`

benchmarking-cli/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+    // Execution time
+    let now = Instant::now();
+    let _trace = miden::execute(&program, &inputs)
+        .map_err(|err| {
+            format!(
+                "Failed to generate exection trace = {:?}, and advice_tape = {:?}",
+                err, advice_tape
+            )
+        })
+        .unwrap();
```
grjte, line 125: I'd prefer `expect` instead of `unwrap`, and if we're not using the result we should be able to get rid of the assignment, right?

grjte (changes requested):
Looks great, thank you! There are just a couple small things I'd prefer to adjust before merging.

benchmarking-cli/src/main.rs
grjte: Actually, I guess this is fine as it is, sorry. I'll mark resolved

benchmarking-cli/src/main.rs
grjte: Actually, the unwrap is fine, sorry. But we should be able to remove the unused variable assignment

benchmarking-cli/src/main.rs
Dominik1999: Actually I need that, `verify` might return an error. So if I remove the `let _result =` then the compiler complains about 

```
unused `Result` that must be used
`#[warn(unused_must_use)]` on by default
this `Result` may be an `Err` variant, which should be handled
```

And if I add an `unwrap()` at the end, it comiles but returns
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "Program failed verification! - VerifierError(InconsistentOodConstraintEvaluations)"', src/main.rs:151:75
```

for the verifiaction step.

benchmarking-cli/src/main.rs
```diff
@@ -0,0 +1,159 @@
+use clap::Parser;
+use miden::{Assembler, ProgramInputs, ProofOptions};
+use winter_math::StarkField;
```
grjte, line 3: I think we'll want to revisit this import in the future to get it from somewhere in Miden instead, but we can leave it this way for now.

grjte (approved):
Looks good, thanks @Dominik1999! I left one comment inline - maybe you can make an issue to address this in the future?

PR #46, opened 2023-01-13
docs: added matrix multiplication example

hackaugusto:
This adds a somewhat complex example, which does matrix multiplication. The examples has code to:

- Load matrix in column or row major order from the advice tape using `adv_pipe` to memory. The format is optimized to stream the data from the tape, and not to be compact. This supports arbitrary sized matrices.
- Code to multiply two matrices, the only limitation on the matrices sizes is that col and row size must be a u32, and the matrices must be compatible to perform the multiplication. The algorithm in cubic.


examples/matrix_multiplication.masm
```diff
@@ -0,0 +1,735 @@
...
+    eqw assert dropw dropw
+    padw push.1001 mem_loadw
+    eqw assert dropw dropw
+    padw push.1000 mem_loadw
+    eqw assert dropw dropw
+
+    dropw dropw dropw
+end
+
+proc.test_matrix_column_major_copy.0
```
hackaugusto, line 698: Testing like this is not ideal, because the later tests will have data later in the advice tape, meaning the advice tape has to be edited. A better approach would be to have tests on separate files, but for an example I think this should be good enough.

examples/matrix_multiplication.masm
```diff
@@ -0,0 +1,735 @@
...
+        # The loop counter `c` will be increasing from 0 to col.
+        # Used for:
+        #   1. End condition when c==col
+        #   2. Offset into addrB
+        #   3. To determine how many padding zeros are needed for row
+        # [[c, r, col, writePtr], [words, readA, addrB, 0], ...]
+        swap
+
+        ## # DEBUG
+        ## dup.0 eq.0 assert         # c==0
```
hackaugusto, line 140: I left debug statements as a sample strategy for loops. It is a bit hard to see what is going on in a loop, specially if the first iteration is successful and a later one fails. In that case testing for code invariants seems to be the best approach.

Initially I started with a `dupw` here, but that made it tricky to understand the number of cycles the assertions would take. This is a problem because at the current version the only reference we have is the number reported in `FailedAssertion`, which makes it hard to figure out which of the assertions failed. This pattern allows all other invariants to be commented out to help figure out what went wrong.

examples/matrix_multiplication.masm
```diff
@@ -0,0 +1,727 @@
...
+    drop drop drop
+end
+
+proc.test_is_zero_or_negative_one.0
+    push.0 exec.is_not_zero_or_negative_one not assert
+    push.0 sub.1 exec.is_not_zero_or_negative_one not assert
+    push.1 exec.is_not_zero_or_negative_one assert
+    push.2 exec.is_not_zero_or_negative_one assert
+    push.0 sub.2 exec.is_not_zero_or_negative_one assert
+end
```
grjte, line 527: I think it would be good to put a separator above this like we do in the miden VM codebase.

Something like the following, with width to column 100
```
# TESTS 
# ==============================================================
```

examples/matrix_multiplication.masm
grjte: Good point, but I agree that it's fine for this example. We can think more about better ways of testing miden assembly as we add more things to the stdlib

examples/matrix_multiplication.masm
```diff
@@ -0,0 +1,727 @@
...
+
+# multiply_add_word(a: Word, b: Word) -> Feld
+#
+# Given two words on the top of the stack, first multiply elements at the same
+# position, then add the results together.
+proc.multiply_add_word.0
+    # input: [A, B, ...]
+    # output: [c, ...]
+
+    # [a1, a2, a3, a4, b1, b2, b3, b4, ...]
```
grjte, line 38: nit: for clarity and consistency with the vm, stdlib, etc, I would write this as `[a3, a2, a1, a0, b3, b2, b1, b0, ...]`. In stack order, the first element of the word is always deepest in the stack, and we also index the elements from zero.

examples/matrix_multiplication.masm
grjte: > In that case testing for code invariants seems to be the best approach.

Agreed. It's a useful pattern - clear, logical, readable.

examples/matrix_multiplication.masm
```diff
@@ -0,0 +1,727 @@
...
+    sub.1
+
+    # [addr, 0, 0, 0, addr, H, H, H, ...]
+    dup movdn.4
+
+    # [row, col, 0, 0, addr, H, H, H, ...]
+    dup sub.1 mem_loadw
+
+    # [row, col, addr, H, H, H, ...]
+    movup.2 drop movup.2 drop
```
grjte, line 471: nit: since this is a duplication of the beginning of the `matrix_row_major_copy`, I probably would have made a helper procedure for the setup. It would be easier to see where these procedures diverge. It's up to you though

grjte (dismissed):
This is really cool, thanks @hackaugusto! I left a few very minor things inline

examples/matrix_multiplication.masm
```diff
@@ -0,0 +1,704 @@
...
+            # [[c, r, colC, writePtr], [words, readA, addrB, 0], acc, ...]
+            swapw add.1 
+
+            # while(c!=colC)
+            dup dup.3 neq
+        end
+
+        # add the padding
+        # [[c, r, colC, writePtr], [words, readA, addrB, 0], acc, ...]
+        dup push.3 u32checked_and neq.0 while.true
```
hackaugusto, line 316: This is the only reason why `u32` is required. I didn't find a better way of figuring out how many padding `0`s should be added to complete the row

examples/matrix_multiplication.masm
```diff
@@ -0,0 +1,704 @@
...
+    # [[rowC, colC, 0, 0], [colA, 0, 0, addrA], addrB, addrC, ...]
+    movup.3 dup.9 mem_storew
+
+    # count the number of words the inner most loop have to consume.
+    # [[words, rowC, colC, 0], [0, 0, 0, addrA], addrB, addrC, ...]
+    movup.4 exec.count_to_word
+
+    # discard extra padding in the middle of the data
+    # [[words, rowC, colC, 0], [0, addrA, addrB, addrC], ...]
+    movup.3 drop movup.3 drop
```
hackaugusto, line 189: operations like this made me wish that drop would accept a bitmask to drop multiple items at once. Here it could be `drop.0b00011`

bobbinth: Yes, that would be cool! Maybe create a feature-request issue for this so we can track it?

hackaugusto: Created :) https://github.com/0xPolygonMiden/miden-vm/issues/642

grjte (approved):
Looks great @hackaugusto, thank you!

Issue #43, opened 2022-12-22
Provide Assembler Instructions 


Dominik1999:
We want to provide Assembler instructions with the Playground. Users should be able to have a nice list handy when they code with the Playground. 

A good example is https://www.evm.codes

## Implementation - done is better than perfect
We will start with a second tab that lists all Assembler instructions. We can improve that in the future. The Assembler Tab will have a searchable list (https://www.npmjs.com/package/react-table) and will cover the same information as our docs. We can think of better ways to show the cycle count in the next iteration. 

## ToDo Lists
- [ ] Create a new tab for the Miden Assembler Instructions
- [ ] Bring all Assembler instructions from our docs into necessary format
- [ ] Show Assembler instructions in react table 
- [ ] Create an issue to improve the Instructions table in the future with automated cycle count

Dominik1999:
@Overcastan Do you know a way how to map our Assembly instructions - [here for example ](https://github.com/0xPolygonMiden/miden-vm/blob/main/docs/src/user_docs/assembly/field_operations.md) - into that format:

```
const instructionSet = [
  {
    name: 'Assertions and tests',
    instructions: [
      { name: 'assert', stack_input: '[a, ...]', stack_output: '[...]', cycles: '1' },
      { name: 'assert_eq', stack_input: '[b, a, ...]', stack_output: '[...]', cycles: '1' },
      { name: '...', stack_input: '...', stack_output: '...', cycles: '...' },
    ],
  },
  // More instructions...
]
```

I think a simple python script would do it.

bobbinth:
I think we should probably spend a bit more time on defining the format, and I'm not sure we should tackle this yet.

For example, one of the issues is that the number of cycles frequently depends on the immediate parameter - so, ideally we should provide number of cycles per instruction variant - but this info is not available in the source file.

Similarly, we can probably do a better job of describing stack start state and end state.

Overcastan:
Indeed, it looks like we can generate such a file using a simple python script, if we allow the use of a variable number of cycles in the description of the operation. But I agree with Bobbin, perhaps we can find a more suitable format. 

I'm not sure yet how we can improve the description of stack states, but the problem of a variable number of cycles is a little clearer to me. If I understand correctly, most of these situations occur when adding an immediate value to the stack, and if it is `1` (or `Felt::ONE`), we use not one operation, but two. It seems to me that we can also handle such cases with a script. 
[...]

PR #47, opened 2023-01-17
Changing the frontend to Tailwind CSS and more

### LINKED ISSUE #7: Refactor the examples playground to use Tailwind CSS instead of MUI
# We want to change to Tailwind CSS instead of MUI 
At the moment in the `App.tsx` we use Material UI. A better choice would be [Tailwind CSS](https://tailwindcss.com/). Tailwind is much easier and more flexible just to use core html + styling (which tailwind makes easy).

Another note: We depend on the entire MUI icons package for one icon. Instead, I would prefer to directly add a new svg component with the Send icon from [HeroIcons](https://heroicons.com/). We can do this when we switch to Tailwind (the same team makes them, and the icons are styled using tailwind)

### LINKED ISSUE #16: Add our logo to the playground
nit: we should add our logo to the icons here

_Originally posted by @grjte in https://github.com/0xPolygonMiden/examples/pull/6#discussion_r1032186834_

### LINKED ISSUE #45: Improve output section in web interface
It would be good to improve the output section of the [web interface](https://0xpolygonmiden.github.io/examples/) to make it more distinguishable and easy to read. The easiest, but not the best way is to simply add something like "stack output" before the output itself. And, maybe, add some separator between output elements: If there are a lot of numbers, and they are large, then it is not always convenient to quickly determine where one number ends and the next number begins. But these are all topics for discussion.

Dominik1999:
* Frontend changed to Tailwind CSS (mui only needed for dropdown), closes #7 
* Added table dummy for Miden assembly instructions, starts #43 
* Added more buttons (debug, prove, verify)
* Output looks better, closes #45
* Added Miden Logo, closes #16


src/App.tsx
```diff
@@ -1,21 +1,14 @@
 import {
```
Dominik1999, line 1: Unfortunately, I still need some MUI elements for the dropdown. I wasn't able to use a Tailwind CSS dropdown that also vanishes when I click somewhere. Also there I have to import a library that uses Tailwind. But maybe one of you have an idea?

src/App.tsx
```diff
@@ -78,107 +67,268 @@ end`
...
...5.2.4 27.3 1 30.6 1.5 7.2 6.9 19 8.7 19 .8 0 3.3-4.3 5.9-9.7 4.3-9.4 4.4-9.8 3.4-15.2-1.3-7.3-1.3-41.2.1-48.7 4.7-25.9 30.6-42.9 56.7-37.3 9.5 2 9.6.8-2.2 25.3-5.8 12-10.6 22-10.6 22.1 0 .2-3.4 7.4-7.6 16.1-4.2 8.8-9.1 19-10.9 22.9-1.8 3.8-6.2 13.1-9.8 20.5-3.6 7.4-8.9 18.7-11.9 25-7.9 16.5-14.4 22.5-27.8 25-10 1.9-17.7-.8-38.7-13.7-7.1-4.4-10.7-8.1-13.1-13.6-1.6-3.6-1.7-129.1-.2-133.3 2.7-7 7.2-10.6 26.5-21.4 2.2-1.2 4.5-2.6 5-3 .6-.4 13.2-7.7 28-16.3 14.9-8.5 33.5-19.2 41.4-23.8 17.3-9.9 20.9-10.8 30.5-7.8zm82.6 47.4c9.3 4.7 24.4 13.7 27.2 16.3 1.5 1.4 3.6 4.5 4.8 6.8 1.9 4.3 2 6 2 67.5 0 45.7-.3 64.3-1.2 67.1-2.2 7.3-6.7 10.6-42.8 31.3-19 10.9-43 24.8-53.5 30.9l-19 11.1h-16l-25.4-14.7-25.3-14.7 6.3-3.2c8.8-4.4 17.7-13.6 22.5-23.3 2-4.2 4-7.6 4.3-7.6.3 0 3.5 1 7.1 2.1 8.5 2.7 23.8 2.9 33 .5 12.6-3.4 24.4-10.7 33.4-20.7 6.3-7 13.3-21.4 15.1-31.2 2.1-11.1 2.1-46.1-.1-56.5-1.7-8.4-5.7-18.2-7.3-18.2-.6 0-3.2 4.4-5.7 9.7l-4.6 9.8.1 25c.2 27-.3 30.4-5.6 40.9-3.3 6.5-12.5 15.7-19.1 19-6.5 3.3-15.2 5.6-21.2 5.6-6.3 0-16.1-2-16.6-3.4-.3-.7 2.9-8.1 7-16.6 11.2-23.6 27.3-57.3 32.6-68.5 2.6-5.5 8-16.8 12-25 4-8.3 8.3-17.5 9.7-20.5 4.9-11.1 11.6-17.9 21.3-21.3 7-2.5 18.1-1.7 25 1.8z" /></svg>
```
Dominik1999, line 72: This is only the Miden Logo

Dominik1999: Jup, I had this before. Would a separate file be better? Isn't the website faster to load when using a picture in this style (might not be super relevant here though).

src/App.tsx
grjte: We should get rid of MUI. I have ideas for this - let's discuss separately

grjte (changes requested):
Looks good! I like the way the inputs and outputs are now. I think beyond that we'll want to get a design anyway, so it works for now.

Regarding the implementation - I left a few notes inline. The main comment I have is that there's a lot of duplication here, and each duplicated container should be pulled out into a component instead that accepts a set of props. I gave an example with some of the buttons, but this is also true of anything that's repeated with the same styling (e.g. the headers for "inputs"/"outputs"/"Miden Assembly Code", the rows/cells of the grid, etc). We can come back through and do a bigger refactor later, but it would be great to do some of it now.

src/App.tsx
```diff
@@ -78,107 +67,268 @@ end`
...
+                let resp = program(code, inputs, numOfOutputs);
+                console.log(resp.toString());
+                setOutput(`Miden VM Program Output \nStack = [${resp.toString()}] \nCycles = <need to return cycles> \n`);
+              } catch (error) {
+                setOutput("Error: Check the developer console for details.");
+              }
+                });
+              }}>
+            Run
+            </button>
```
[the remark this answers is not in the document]
grjte: You can look up presentational vs. container components in React to learn more about this design pattern

src/App.tsx
```diff
@@ -78,107 +67,268 @@ end`
...
+
+            <button className="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded">
+            Debug
+            </button>
+            <button className="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded">
+            Prove
+            </button>
+            <button className="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded">
+            Verify
+            </button>
```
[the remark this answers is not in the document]
Dominik1999: Ah got it, yes that is way cleaner. Let me try to apply this pattern on more components

src/App.tsx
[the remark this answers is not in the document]
Dominik1999: nice, that makes more sense and it makes everything cleaner

src/App.tsx
Dominik1999: perfect, done

src/App.tsx
Dominik1999: No more MUI.

src/App.tsx
```diff
@@ -60,10 +57,16 @@ end`
...
+  * If a new example is selected using the dropdown, the code and inputs are updated.
+  */  
+
   const [example, setExample] = React.useState<string>();
-  const handleSelectChange = async (event: SelectChangeEvent) => {
-    const value = event.target.value;
+  console.log(example)
+  const handleSelectChange = async (exampleChange: string) => {
+    const value = exampleChange;
+    console.log("adad")
```
grjte, line 69: Should we remove these console logs?

src/App.tsx
```diff
@@ -76,109 +79,96 @@ end`
...
+            <div className="text-sm lg:flex-grow">
+              <a href="https://wiki.polygon.technology/docs/miden/user_docs/assembly/main" className="block mt-4 lg:inline-block lg:mt-0 text-teal-200 hover:text-white mr-4">
+                Docs
+              </a>
+              <a href="https://github.com/0xPolygonMiden/examples#available-examples" className="block mt-4 lg:inline-block lg:mt-0 text-teal-200 hover:text-white mr-4" target="_blank" rel="noopener noreferrer">
+                README: Examples
+              </a>
+              <a href="https://polygon.technology/solutions/polygon-miden/" className="block mt-4 lg:inline-block lg:mt-0 text-teal-200 hover:text-white">
+                Homepage
+              </a>
```
grjte, line 116: I would probably create a `Link` component or at least pull the custom styling out into a `globals.css` file so it's not duplicated. https://tailwindcss.com/docs/adding-custom-styles#adding-component-classes

src/components/ActionButton.tsx
```diff
@@ -0,0 +1,18 @@
+type ActionButtonProps = {
+    label: string,
+    onClick?: any,
```
grjte, line 3: nit: ideally, we should avoid using "any" types. You should be able to use the following:

```
onClick?: () => void;
```

src/components/DropDown.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return classes.filter(Boolean).join(' ')
+}
+
+type DropDownProps = {
+    onExampleValueChange?: (newType: string) => void;
+}
+
+export default function DropDown({ onExampleValueChange }: DropDownProps) {
+  const [selected, setSelected] = useState(examples[1])
+  //console.log(selected.name)
```
grjte, line 24: We can probably remove this

src/components/DropDown.tsx
```diff
@@ -0,0 +1,88 @@
...
+            </Transition>
+          </div>
+        </>
+      )}
+    </Listbox>
+  )
+}
+function setTags(arg0: string[]) {
+  throw new Error('Function not implemented.')
+}
```
grjte, line 87: do we need this?

package.json
```diff
@@ -5,14 +5,17 @@
   "private": true,
   "dependencies": {
     "@codemirror/theme-one-dark": "^6.1.0",
+    "@headlessui/react": "^1.7.7",
+    "@heroicons/react": "^2.0.13",
```
grjte, line 9: Since we're not using very many icons, we could just copy the code for the few that we need from the site and remove this dependency. https://heroicons.com/

src/components/InstructionTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+                    <th scope="col" className="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">
+                      Stack Input
+                    </th>
+                    <th scope="col" className="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">
+                      Stack Output
+                    </th>
+                    <th scope="col" className="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">
+                      Cycles
+                    </th>
+                    <th scope="col" className="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">
```
grjte, line 48: I would define a named for "px-3 py-3.5 text-left text-sm font-semibold text-gray-900" in `globals.css` so it doesn't need to be duplicated

src/App.tsx
```diff
@@ -1,57 +1,54 @@
...
 
-  const [numOfOutputs, setNumOfOutputs] = React.useState(1);
+  const [numOfOutputs, setNumOfOutputs] = React.useState(16);
 
-  // this should be dyncamically read from ./examples/ - buy my TS knowledge is limited
-  const examples = [
+  /**
+  * This lists all the examples available in the examples folder.
+  */  
+ const examples = [
```
grjte, line 51: Do we still need this defined here? It's also defined in the `Dropdown` component.

src/App.tsx
```diff
@@ -1,57 +1,54 @@
...
+
 
 async function getExample(example: string) {
     const inputs = fetch(`https://raw.githubusercontent.com/0xPolygonMiden/examples/main/examples/${example}.inputs`)
     const masm = fetch(`https://raw.githubusercontent.com/0xPolygonMiden/examples/main/examples/${example}.masm`)
     return [(await inputs).text(), (await masm).text()];
   }
 
-function App() {
+function App(this: any) {
```
grjte, line 18: why do we need this `this: any`?

grjte (changes requested):
Thanks @Dominik1999 this refactor looks great! I left a few small comments inline. Aside from those, the final thing I'd recommend changing is to reduce the duplication in the styles. We can define custom styles with tailwind so components/tags that share the same 5 style items can each get one shared style name instead. I left a couple of related comments inline, and you can learn more here: https://tailwindcss.com/docs/adding-custom-styles#adding-component-classes

Feel free to ping me directly as well

src/components/DropDown.tsx
Dominik1999: nope

src/App.tsx
Dominik1999: nope its a left over

src/components/InstructionTable.tsx
Dominik1999: done

package.json
Dominik1999: you are right

grjte (changes requested):
Looks great now! I just noticed that there's still a dependency for the `@mui/icons-material`. Do we still need this? Also, I would have expected the `package-lock.json` to change after you removed the dependency on the heroicons, but it didn't.

Can you just double check on the dependencies to make sure we've cleaned everything up? After that I think we're ready to merge

Dominik1999:
> Looks great now! I just noticed that there's still a dependency for the `@mui/icons-material`. Do we still need this? Also, I would have expected the `package-lock.json` to change after you removed the dependency on the heroicons, but it didn't.
> 
> Can you just double check on the dependencies to make sure we've cleaned everything up? After that I think we're ready to merge

Perfect, I removed the dependent and another file. I think we can merge.

grjte (approved):
Yep, good catch with the `theme.ts` file as well. Looks good, thanks!

PR #49, opened 2023-01-25
refactor: change repo structure

### LINKED ISSUE #11: Change repo structure and put playground into own folder

Dominik1999:
Closes #11 

Changes Repo Structure and puts the playground into its folder. 

I think it still gets build with gh pages, let's see


.gitignore
```diff
@@ -1,15 +1 @@
...
-/node_modules
-
-# testing
-/coverage
-
-# production
-/build
-
-# misc
-.DS_Store
```
grjte, line 11: We could probably keep this one. Also, I like having the little section comments to keep the .gitignore organized, since we'll probably add things in the future, so I'd keep the ones for `# dependencies` and `# misc`

playground/tsconfig.json
```diff
@@ -17,5 +17,5 @@
     "jsx": "react-jsx",
     "types": [  "node"  ]
   },
-  "include": ["src"],
+  "include": ["playground/src"],
```
grjte, line 20: Does this need to change?

.github/workflows/gh-pages.yml
```diff
@@ -21,11 +21,11 @@ jobs:
       - name: Install and Build 🔧
         run: |
           npm install
-          npm run preinstall
+          npm run build:miden
```
grjte, line 24: I don't think this will work, but we can test it by creating a dummy branch (e.g. `dominik-test-deployment`) and updating the `gh-pages.yml` file on the dummy branch to:

```
name: Deploy Miden Assembly Playground
on:
  push:
    branches:
      - dominik-test-deployment

...
```

README.md
```diff
@@ -41,10 +41,10 @@ You can also run the playground locally on your machine.
 
 If you want to run it locally, then make sure you have the [wasm-pack](https://rustwasm.github.io/wasm-pack/installer/) installed.
 
-To run the playground locally in development mode, navigate to the project directory and run:
+To run the playground locally in development mode, navigate to the project directory `playground` and run:
 
 ```
-npm install && npm start
+npm install && npm run build:miden && npm run start
```
grjte, line 47: This should be changed to the following in order to work (otherwise `npm install` will fail):

`npm run build:miden && npm install && npm run start`

grjte (changes requested):
Thanks @Dominik1999! I left a few comments inline

.gitignore
Dominik1999: Yes, indeed. This was just in the wrong place. We need the .gitignore in the /playground folder

.github/workflows/gh-pages.yml
Dominik1999: good point. Let's try. I do still not understand how exactly I can test this, buy now I will get into the weeds

playground/tsconfig.json
Dominik1999: I don't think so

.github/workflows/gh-pages.yml
Dominik1999: I am not sure how I can start the deployment tbh. I must be merged for that, right?

.github/workflows/gh-pages.yml
```diff
@@ -21,11 +21,11 @@ jobs:
...
           npm run build
           npx webpack
       - name: Deploy to GitHub Pages
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: build
\ No newline at end of file
+          publish_dir: build
```
grjte: I would keep the newline at the end of this file

.github/workflows/gh-pages.yml
Dominik1999: ok now it works

.github/workflows/gh-pages.yml
Dominik1999: done

.github/workflows/gh-pages.yml
```diff
@@ -20,12 +20,13 @@ jobs:
...
```
grjte: Let's keep the newline at the end of the file

grjte (changes requested):
Thanks @Dominik1999! Can you fix the comment I left inline and squash out the test commits before we merge?

.github/workflows/gh-pages.yml
Dominik1999: done

grjte (approved):
Looks good, thank you!

Issue #18, opened 2022-11-25
Catch any input error from the frontend


Dominik1999:
Currently, we get errors when the input is not correctly formatted, or `stack_init` is missing. We need to catch any input before deserialising it in the Rust backend. We can have a pop-up window warning the user that a certain input cannot be accepted. We can use https://react-hot-toast.com/ to prompt a pop-up.

### Input data
Currently, we support `stack_init` and `advice_tape` as inputs. 

The playground should accept any input that:
- is empty
- only consists of `stack_init` and is a valid JSON, and `stack_init` consists of at least one integer and only consists of integers
- only consists of `advice_tape` and is a valid JSON, and `advice_tape` consists of at least one integer and only consists of integers
- consists of `stack_init` and `advice_tape` and is a valid JSON, and `stack_init` and `advice_tape` consists of at least one integer and only consist of integers

### Input code
Currently, we support any code input. There is no restriction on valid Assembly code. But we should check that the code does not go over a certain length.

Dominik1999:
https://react-hot-toast.com/

Dominik1999:
@grjte @Overcastan 

Any thoughts on the above? What do we want to allow as inputs? A clear definition could help the user. 

Miden already rejects malformatted input anyway. In Miden, the `InputFile` is read here 

https://github.com/0xPolygonMiden/miden-vm/blob/9c8af67a7768b920772b3a42d89947f7f639e089/miden/src/cli/data.rs#L16-L26

But I could not find a straight definition of what an input should look like.

PR #50, opened 2023-01-26
feat: Proving for the playground

### LINKED ISSUE #19: Add ability to prove a program as well
At the moment, we only execute the Miden assembly program. We should add the ability to prove and also to verify a program.

### Proving a Miden program in lib.rs
We can add the feature as another function, `pub fn prove_program` and also have the program, the inputs and the output_count as function inputs. 

The problem so far is, that we can't return a tuple or a complex struct from Rust to the frontend using `wasm-bindgen`, see https://stackoverflow.com/questions/75064800/how-to-return-a-tuple-to-frontend-from-wasm

There are solutions, but I was not able to implement them in #50 

Another problem is that there is no info about the trace or the time exposed in Miden v0.3. So we don't know how many cycles or how long it took, see https://github.com/0xPolygonMiden/examples/issues/29. The problem with the duration of the proving step could be solved by adding a timestamp to the frontend. 

### Proving a Miden program in the App.tsx
We add another function like for `run_program`. The return values so far are the stack_output and the overflow_addresses. 
However, we do not get the proof or any other info yet to print, see above.

In an ideal world, we would get some info while the proving step is executed, like the info that is logged when you prove via the CLI.

### Open questions to clarify in the working group
Maybe the ability to prove is enough for this iteration, but if not we need to answer 

- What do we want to output while proving?
- What do we want to output during proving?
- How do we want to show the proof?

Any opinions / ideas on the above @grjte or @Overcastan ?

Dominik1999:
Closes #19 

This PR enables proving to the playground. 

There are some open questions regarding the UX and the proving step code itself:

- UX
- - At the moment, for the user running and proving a program looks equal. The output is almost the same. 
- - We should add some info that proving might take a while (see Game of Life example), maybe even a pop-up that prints something during the proof
- - There is no way to show the proof atm

- Proving step
- - We don't expose in Miden v0.3 any info about the trace length or cycles or even time to prove in `ProgramOutputs` or `ExecutionTrace`, so we can't get the numbers (maybe I am mistaken, but I didn't find it)
- - I couldn't find a way to pass the proof from Rust to Typescript (frontend) via wasm-bindgen. There seems to be a way to return complex structures from Rust to Typescript (frontend) but I didn't understand exactly how

### DIFF
--- a/playground/miden-wasm/src/lib.rs
+++ b/playground/miden-wasm/src/lib.rs
@@ -1,4 +1,4 @@
-use miden::{Assembler, ProgramInputs};
+use miden::{Assembler, ProgramInputs, ProofOptions};
 use wasm_bindgen::prelude::*;
 use wasm_bindgen_test::*;
 extern crate console_error_panic_hook;
@@ -36,7 +36,7 @@ fn get_advice_tape(inputs: &InputFile) -> Vec<u64> {
 
 /// Runs the Miden VM with the given inputs
 #[wasm_bindgen]
-pub fn program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64> {
+pub fn run_program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64> {
     console_error_panic_hook::set_once();
     wasm_logger::init(wasm_logger::Config::default());
 
@@ -63,16 +63,62 @@ pub fn program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64> 
     )
     .map_err(|err| format!("Failed to generate exection trace = {:?}, and advice_tape = {:?}", err, advice_tape)).unwrap();
     
-    // TODO: Investigate why proof verification fails when outputCount > 1
-    //assert!(miden::verify(*program.hash(), &[], &outputs, proof).is_ok());
-    // tracked in https://github.com/0xPolygonMiden/examples/issues/19
     trace.program_outputs().stack_outputs(output_count as usize).to_vec()
 }
 
-/// Basic test for the program
+/// Proves the program with the given inputs
+#[wasm_bindgen]
+pub fn prove_program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64> {
+    console_error_panic_hook::set_once();
+    wasm_logger::init(wasm_logger::Config::default());
+
+    let assembler = Assembler::new();
+    let program = assembler.compile(&asm).expect("Could not compile source");
+    
+    let mut stack_init = <Vec<u64>>::new();
+    let mut advice_tape = <Vec<u64>>::new();
+
+    // TODO: We must catch all cases of inputs. Input can be not a valid json,
+    // input can be empty, stack_init can be empty, advice_tape can be empty. And all combinations.
+    if inputs_frontend.trim().is_empty() == false {
+        let inputs_des: InputFile = serde_json::from_str(&inputs_frontend)
+        .map_err(|err| format!("Failed to deserialize input data - {}", err)).unwrap();
+        
+        stack_init = get_stack_init(&inputs_des);
+        advice_tape = get_advice_tape(&inputs_des);
+    } 
+
+    let inputs = get_program_inputs(&stack_init, &advice_tape); 
+
+    // default (96 bits of security)
+    let proof_options = ProofOptions::with_96_bit_security();
+
+    let (output, _proof) = miden::prove(
+        &program,
+        &inputs,
+        &proof_options,
+    )
+    .map_err(|err| format!("Failed to generate exection trace = {:?}, and advice_tape = {:?}", err, advice_tape)).unwrap();
+    
+    // We convert the outputs into one Vector because we can't return tuples in wasm-bindgen.
+    // So the first n elements are the stack_outputs and the last n elements are the overflow_addrs.
+    // TODO: We can somehow refactor to return structures in wasm-bindgen. But I didn't find a way yet.
+    // There might be a way to return the ProgramOutputs struct even.
+    let mut stack_outputs = output.stack_outputs(output_count as usize).to_vec();
+    let mut overflow_addrs = output.overflow_addrs().to_vec();
+    stack_outputs.append(&mut overflow_addrs);
+    
+    // TODO: We must return trace_info as well, but we can not handle that yet in Miden v0.3.
+    // TODO: We must return the proof as well. But we can not handle that yet in the frontend.
+
+    stack_outputs
+
+}
+
+/// Basic tests for the program
 #[wasm_bindgen_test]
-fn run_program() {
-    let output = program(
+fn test_run_program() {
+    let output = run_program(
         "begin
             push.1 push.2 add
         end",
@@ -80,4 +126,21 @@ fn run_program() {
         1,
     );
     assert_eq!(output[0], 3)
+}
+
+#[wasm_bindgen_test]
+fn test_prove_program() {
+    let output = prove_program(
+        "begin
+            push.1 push.2 add
+        end",
+        "",
+        16,
+    );
+    // this is the result of the stack output, 3
+    assert_eq!(output[0], 3);
+
+    // for the proof we have [0, 1] as overflow_addrs
+    assert_eq!(output[16], 0);
+    assert_eq!(output[17], 1);
 }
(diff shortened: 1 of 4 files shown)


grjte:
Since this will need to be rebased and updated after the repo structure PR is changed, let's keep it as a draft until then. I'm going to put it into draft mode and remove my review request. Once it's been rebased, please mark it as "ready for review" and then re-request review

grjte:
> There are some open questions regarding the UX and the proving step code itself:
> 
>     * UX
> 
>     * * At the moment, for the user running and proving a program looks equal. The output is almost the same.

I think this is addressed by discussions below. As long as we have a way to provide/show the proof then it's fine for the output to be the same (imo). Is there something else you were thinking about here?

>     * * We should add some info that proving might take a while (see Game of Life example), maybe even a pop-up that prints something during the proof

We can use react-hot-toast to do this very easily: https://react-hot-toast.com/

> 
>     * * There is no way to show the proof atm

I think we should create a new issue for this, but reference #19 inside that issue, as well as the other related open questions/issues you've noted

>     * Proving step
> 
>     * * We don't expose in Miden v0.3 any info about the trace length or cycles or even time to prove in `ProgramOutputs` or `ExecutionTrace`, so we can't get the numbers (maybe I am mistaken, but I didn't find it)

I think this is part of a larger conversation in the VM, and related to changes for debugging improvements as well (https://github.com/0xPolygonMiden/miden-vm/issues/580). We should bring it up on the team sync call, but maybe this should be separated out from this set of playground chanages and become its own working group that looks at the 2 things together.

>     * * I couldn't find a way to pass the proof from Rust to Typescript (frontend) via wasm-bindgen. There seems to be a way to return complex structures from Rust to Typescript (frontend) but I didn't understand exactly how

This seems to be related to the comment above about having a way to show the proof, so it seems to me that we should just create a new issue for showing the proof.

Dominik1999:
> Since this will need to be rebased and updated after the repo structure PR is changed, let's keep it as a draft until then. I'm going to put it into draft mode and remove my review request. Once it's been rebased, please mark it as "ready for review" and then re-request review

Thanks, I just rebased.

Dominik1999:
> > ```
> > * * I couldn't find a way to pass the proof from Rust to Typescript (frontend) via wasm-bindgen. There seems to be a way to return complex structures from Rust to Typescript (frontend) but I didn't understand exactly how
> > ```
> 
> This seems to be related to the comment above about having a way to show the proof, so it seems to me that we should just create a new issue for showing the proof.

Here is a new issue about getting and showing the proof. We can tackle this in the next milestone. https://github.com/0xPolygonMiden/examples/issues/52

grjte:
> Here is a new issue about getting and showing the proof. We can tackle this in the next milestone. #52

Sounds good! I think we should also create an issue about adding toast messages and put that in the "nice to have" for this milestone. wdyt?

playground/miden-wasm/src/lib.rs
```diff
@@ -63,16 +63,61 @@ pub fn program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64>
...
+    // We convert the outputs into one Vector because we can't return tuples in wasm-bindgen.
+    // So the first n elements are the stack_outputs and the last n elements are the overflow_addrs.
+    // TODO: We can somehow refactor to return structures in wasm-bindgen. But I didn't find a way yet.
+    // There might be a way to return the ProgramOutputs struct even.
+    let mut stack_outputs = output.stack_outputs(output_count as usize).to_vec();
+    let mut overflow_addrs = output.overflow_addrs().to_vec();
+    stack_outputs.append(&mut overflow_addrs);
+    
+    // TODO: We must return trace_info as well, but we can not handle that yet in Miden v0.3.
+    // TODO: We must return the proof as well. But we can not handle that yet in the frontend.
```
grjte, line 111: Let's just make sure these are both captured in an issue (I think they are, but can you confirm?)

playground/miden-wasm/src/lib.rs
```diff
@@ -63,16 +63,61 @@ pub fn program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64>
...
+    console_error_panic_hook::set_once();
+    wasm_logger::init(wasm_logger::Config::default());
+
+    let assembler = Assembler::new();
+    let program = assembler.compile(&asm).expect("Could not compile source");
+    
+    let mut stack_init = <Vec<u64>>::new();
+    let mut advice_tape = <Vec<u64>>::new();
+
+    // TODO: We must catch all cases of inputs. But maybe in the frontend already.
```
grjte, line 81: I don't understand this. Can you rephrase and also make sure it's tracked by an issue?

playground/miden-wasm/src/lib.rs
```diff
@@ -63,16 +63,61 @@ pub fn program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64>
...
+        &program,
+        &inputs,
+        &proof_options,
+    )
+    .map_err(|err| format!("Failed to generate exection trace = {:?}, and advice_tape = {:?}", err, advice_tape)).unwrap();
+    
+    // We convert the outputs into one Vector because we can't return tuples in wasm-bindgen.
+    // So the first n elements are the stack_outputs and the last n elements are the overflow_addrs.
+    // TODO: We can somehow refactor to return structures in wasm-bindgen. But I didn't find a way yet.
+    // There might be a way to return the ProgramOutputs struct even.
```
grjte, line 105: We can leave this for another PR. Let's just make sure it's tracked in an  issue

playground/miden-wasm/src/lib.rs
```diff
@@ -63,16 +63,61 @@ pub fn program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64>
...
+    stack_outputs
+
+}
+
 /// Basic test for the program
 #[wasm_bindgen_test]
-fn run_program() {
-    let output = program(
+fn test_run_program() {
+    let output = run_program(
```
grjte, line 120: Does it make sense to add a basic test for proving too?

grjte (changes requested):
Looks great, thank you! I left a few very minor comments. Also, I want to be sure all inline TODOs are captured in issues. I think many/most are, but just to check

playground/miden-wasm/src/lib.rs
Dominik1999: Yes, it's tracked in https://github.com/0xPolygonMiden/examples/issues/52

playground/miden-wasm/src/lib.rs
Dominik1999: this is also tracked in https://github.com/0xPolygonMiden/examples/issues/52

playground/miden-wasm/src/lib.rs
Dominik1999: This is tracked here - and the next PR that I will send to you https://github.com/0xPolygonMiden/examples/issues/18

I will rephrase

playground/miden-wasm/src/lib.rs
Dominik1999: Hmm indeed, we should add that.

Dominik1999: added a test

playground/miden-wasm/src/lib.rs
```diff
@@ -63,16 +63,62 @@ pub fn program(asm: &str, inputs_frontend: &str, output_count: u16) -> Vec<u64>
...
+    wasm_logger::init(wasm_logger::Config::default());
+
+    let assembler = Assembler::new();
+    let program = assembler.compile(&asm).expect("Could not compile source");
+    
+    let mut stack_init = <Vec<u64>>::new();
+    let mut advice_tape = <Vec<u64>>::new();
+
+    // TODO: We must catch all cases of inputs. Input can be not a valid json,
+    // input can be empty, stack_init can be empty, advice_tape can be empty. And all combinations.
```
grjte, line 82: nit: I see - you're talking about cases of input errors. I would probably have said "all cases of input errors", but the message is now clear, and no need to update, especially since it'll be tackled in the next PR.

grjte (approved):
Approved! Thank you!

playground/miden-wasm/src/lib.rs
Dominik1999: ah indeed. I like it those nits

PR #51, opened 2023-01-26
fix: pages build but do not show

Dominik1999:
The playground homepage doesn't show, it can't find index.html. I think I fixed it now


playground/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "miden-assembly-examples-playground",
   "version": "0.1.0",
-  "homepage": "https://0xpolygonmiden.github.io/examples/",
+  "homepage": "https://0xpolygonmiden.github.io/examples/playground/",
```
grjte, line 4: If this is being changed, then we should also update the playground link in the `README` (and anywhere else)

grjte (changes requested):

playground/package.json
Dominik1999: Indeed. But I don't know if the change in the workflow will also change the URL. Somehow this is still a black box to me. Actually, it looks like the URL stays the same, `yourusername.github.io/your_repo_name` according to yourusername.github.io/your_repo_name

Dominik1999: But to be honest, that is a bit tedious trial and error (for me)

grjte (approved):
Looks good, thanks!

PR #53, opened 2023-02-01
feat: introduce shamir secret share example

frisitano:
This PR introduces an example for creating shares in a shamir secret sharing scheme.


examples/shamir-secret-share.masm
```diff
@@ -0,0 +1,150 @@
...
+        lt
+    end         #[k, counter, coef_1, ...]  
+    
+    # evaluate if we have too many cofficients
+    dup          #[k, k, counter, coef_1, ...]
+    movdn.2      #[k, counter, k, coef_1, ...]
+    sub          #[diff, k, coef_1, ...]
+    dup          #[diff, diff, k, coef_1, ...]
+    push.0       #[0, diff, diff, k, coef_1, ...]
+    gt           #[1, diff, k, coef_1]
```
hackaugusto, line 80: mega nit: I don't think it is necessary to rewrite this, but it may be cool to share this strategy. Since the loop above produces `4` coefficients per iteration, it would be nice if `k` was a multiple of it. That can be done with something like:

```
               # stack: [e, ...]
u32cast        # stack: [low, ...] discard high bits
push.3         # stack: [4, low, ...] bit pattern for the lower bits
u32checked_and # stack: [diff, ...] the remainder
```

It is just `3` cycles to get the remainder. With it you can compute the number of loops:

```
               # stack: [k, ...]
dup.0          # stack: [k, k, ...]
exec.remainder # stack: [r, k, ...]
dup.1 dup.1    # stack: [r, k, r, k, ...]
sub            # stack: [loop, r, k, ...]
```

Later you set counter to `0` and increment it by `4` until it is equal to `loop`. This can be done with a `neq` check which is one cycle, instead of the `lt` which is `17`.

The remainder can be handle on a similar fashion, first get the number of elements to discard `4-r`, then set a counter variable to `0` and in the loop 1. drop the extra coef 2. increment the counter 3. do a `neq` check. This removes the `gt` which is `18` cycles in favor of a `neq` which is only one!

I think it is a neat trick to optimize cycle count :)

examples/shamir-secret-share.masm
```diff
@@ -0,0 +1,150 @@
...
+    #[x, k, coef_1,...]
+
+    # introduce sum variable to calculate y
+    swap    #[k, x, coef_1, ..., secret]
+    push.0  #[sum, k, x, coef_1, ..., secret]
+    push.1  #[1, sum, k, x, coef_1, ..., secret]
+
+    # evaluate polymoial by iterating through terms and summing
+    while.true 
+        dup.2 #[x, sum, k, x, coef_1,..., secret]
```
hackaugusto, line 106: on the first iteration there is a `1` on the top of the stack, so this is really:

```suggestion
        dup.2 #[k, 1, sum, k, x, coef_1,..., secret]
```

I didn't review the rest of the loop, it seems the `1` is not accounted for the ops below, so the stack should be reviewed.

hackaugusto (changes requested):
Awesome! I did a first pass, left some nits and some comments that need a second look, maybe they are bugs or I made some mistakes ;)

examples/shamir-secret-share.masm
```diff
@@ -0,0 +1,150 @@
...
+#!
+#! Input:
+#! Stack: [x, k] 
+#! The x coordinate for which we want to generate a share, number of shares  needed to 
+#! reconstruct secret - this defines the degree of the polynomial.
+#! Advice Tape: [secret, rand]
+#! The secret, a random field element.
+#!
+#! Output:
+#! Stack: [x_coord, y_coord, com_1, com_2, com_3, com_4]
```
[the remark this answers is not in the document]
grjte: I would also prefer this change :)

examples/shamir-secret-share.masm
```diff
@@ -0,0 +1,150 @@
...
+    # continue creating more coefficients until we have enough (created in 4's)
+    while.true     #[k, counter, diget_1, diget_2, digest_3, digest_4]
+        # generate more cofficients by padding and hashing previous cofficients
+        movdn.5
+        movdn.5
+        dupw       #[digest_1, digest_2, digest_3, digest_4, k, counter, ...]
+        padw       #[0, 0, 0, 0, diget_1, diget_2, digest_3, digest_4, diget_1, diget_2, digest_3, digest_4, k, counter]
+        rphash     #[digest_5, digest_6, digest_7, digest_8, digest_1, digest_2, digest_3, digest_4, k, counter, ...]
+
+        # evaluate if we have enugh coeficients
```
grjte, line 65: nit: typos in "enough" and "coefficients"

examples/shamir-secret-share.masm
```diff
@@ -0,0 +1,150 @@
...
+# evaluate polynomial
+proc.eval_poly
+    #[x, k, coef_1,...]
+
+    # introduce sum variable to calculate y
+    swap    #[k, x, coef_1, ..., secret]
+    push.0  #[sum, k, x, coef_1, ..., secret]
+    push.1  #[1, sum, k, x, coef_1, ..., secret]
+
+    # evaluate polymoial by iterating through terms and summing
```
grjte, line 104: nit: typo in "polynomial"

examples/shamir-secret-share.masm
```diff
@@ -0,0 +1,150 @@
...
+        # evaluate if we have enugh coeficients
+        movup.9
+        add.4
+        movup.9
+        dup.1
+        dup.1
+        lt
+    end         #[k, counter, coef_1, ...]  
+    
+    # evaluate if we have too many cofficients
```
grjte, line 74: nit: typo in "coefficients"

grjte (commented):
Great pick, thank you! I haven't left any substantial comments, since Augusto covered most of it and I think there may be some changes based on what he's said, so I just left a few nits inline.

One other comment is that the costs are unexpected sometimes, because many of these assembly operations are combined of several vm instructions under the hood. For example, `lt` and `gt` are expensive, so whenever it's possible to do something else, that's usually preferable. 

In addition to the costs noted in the docs (which include ranges instead of exact numbers), you can look here in the code to see how things are implemented, as well as the specific parser methods for those operations that have them.
https://github.com/0xPolygonMiden/miden-vm/blob/main/assembly/src/assembler/instruction/mod.rs

frisitano:
> Awesome! I did a first pass, left some nits and some comments that need a second look, maybe they are bugs or I made some mistakes ;)

Hey, thanks for the review!  A nice set of tips / conventions that I'll be taking onboard.

I don't think there are any logical bugs as I am able to reconstruct the secret.  Let's do an example:

k = 3
secret = 12345678
rand = 10

k = 3 and therefore we need three sets of coords to reconstruct the secret:

x = 435956367654746 
result = [435956367654746, 5512490206371841177, 10162399191635666462, 8066465782393336362, 13516526422452661581, 1550228670226960766]

x = 25676367654746
result = [25676367654746, 5717940160797348684, 10162399191635666462, 8066465782393336362, 13516526422452661581, 1550228670226960766]

x = 12324352355746
result = [12324352355746, 7917744552196057384, 10162399191635666462, 8066465782393336362, 13516526422452661581, 1550228670226960766]

Lets plug the (x, y) coordinates into sage maths and interpolate over the field:

Screenshot from 2023-02-02 09-22-03

examples/shamir-secret-share.masm
```diff
@@ -0,0 +1,150 @@
...
+
+    # generate initial 4 random coefficients from rand
+    swap    #[rand, k]
+    padw    #[0, 0, 0, 0, rand, k]
+    padw    #[0, 0, 0, 0, 0, 0, 0, 0, rand, k]
+    drop    #[0, 0, 0, 0, 0, 0, 0, rand, k]
+    rphash  #[coef_1, coef_2, coef_3, coef_4, k]
+
+    # introduce counter and evaluate if we have enough cofficients
+    push.5  #[counter, coef_1, coef_2, coef_3, coef_4, k]
```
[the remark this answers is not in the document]
frisitano: It needs to be 5 as we also need to take into account the secret which acts as the intercept.  I will add a comment to make this clearer.

examples/shamir-secret-share.masm
frisitano: Awesome trick, thank you for sharing!  When I get some free time I will implement this as it'll be a good way to practice my assembly optimisation.

examples/shamir-secret-share.masm
frisitano: I think the 1 is consumed by the while loop condition evaluation.

examples/shamir-secret-share.masm
```diff
@@ -0,0 +1,150 @@
+#! This program creates verifiable shares for a shamir secret sharing scheme.
+#! This involves generation of random polynomial coefficients and 
+#! polynomial evaluation in the prime field.  The dealer generates shares 
+#! independently such that they can be distributed independently, the same
+#! secret, random seed and k value must be used to create all shares.
```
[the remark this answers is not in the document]
frisitano: good catch.

examples/shamir-secret-share.masm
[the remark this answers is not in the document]
frisitano: This make sense.  Thanks for providing the reference @hackaugusto.

frisitano:
Thanks both for the review.  I've addressed the points raised.  Something to note is that when generating coefficients, they start from `coef_1` as `coef_0` is the secret.  I'm not sure if it's best to implement the optimisation @hackaugusto suggested before we get this merged or if it's best to leave that for a follow up activity.   What are your thoughts?

grjte:
> Thanks both for the review. I've addressed the points raised. Something to note is that when generating coefficients, they start from `coef_1` as `coef_0` is the secret. I'm not sure if it's best to implement the optimisation @hackaugusto suggested before we get this merged or if it's best to leave that for a follow up activity. What are your thoughts?

This whole thing is kind of just for fun anyway & to get rolling with Miden assembly, so there's no need to implement the optimization unless you want to. That said, it may be unlikely that we'll follow up on this, at least not in the near future, so if you feel like doing it while things are fresh, go for it. Either way is fine.

frisitano:
> This whole thing is kind of just for fun anyway & to get rolling with Miden assembly, so there's no need to implement the optimization unless you want to. That said, it may be unlikely that we'll follow up on this, at least not in the near future, so if you feel like doing it while things are fresh, go for it. Either way is fine.

Ok in that case I think we should just leave it as is.  I think the objective was achieved.  I became familiar with Miden assembly and learnt some lessons along the way.

examples/shamir-secret-share.masm
```diff
@@ -0,0 +1,150 @@
...
+    swap    #[k, x, coef_k-1, ..., secret]
+    push.0  #[sum, k, x, coef_k-1, ..., secret]
+    push.1  #[1, sum, k, x, coef_k-1, ..., secret]
+
+    # evaluate polynomial by iterating through terms and summing
+    while.true 
+        dup.2 #[x, sum, k, x, coef_k-1,..., secret]
+        dup.2 #[k, x, sum, k, x, coef_k-1,..., secret]
+        sub.1 #[k-1, x, sum, k, x, coef_k-1,..., secret]
+        exp   #[x^k-1, sum, k, x, coef_k-1,..., secret]
```
grjte, line 109: nit: `x^(k-1)` here and below. It might be clearer to use parens for `coef_(k-1)` as well

grjte (approved):
Looks great, thanks again @frisitano for a very cool example! I left one non-blocking nit inline, but I'll leave the merge to you so you can update it if you want to.

grjte:
@hackaugusto I think we're blocked from merging until you approve. Can you take another look at this when you get a chance?

hackaugusto (approved):
Approving the PR, seems all the comments have been addressed.

Issue #28, opened 2022-11-30
Build reasonable testing for the playground


Dominik1999:
At the moment, the playground is tested poorly. We need to test the Typescript Frontend and the WASM backend.

The correctness of the examples will be tested in a separate issue. #56 

- [x] Test the frontend using https://reactjs.org/docs/testing.html
- [x] Test the WASM part using unit tests and one integration test

Dominik1999:
Closed by #66

PR #54, opened 2023-02-06
feat: catching any input errors in the frontend

### LINKED ISSUE #18: Catch any input error from the frontend

Dominik1999:
Closes #18

### DIFF
--- a/playground/src/pages/CodingEnvironment.tsx
+++ b/playground/src/pages/CodingEnvironment.tsx
@@ -5,6 +5,7 @@ import ActionButton from "../components/ActionButton"
 import DropDown from "../components/DropDown"
 import CodeMirror from "@uiw/react-codemirror";
 import init, { run_program, prove_program } from "miden-wasm";
+import toast, { Toaster } from 'react-hot-toast';
 
 
 async function getExample(example: string) {
@@ -14,6 +15,96 @@ async function getExample(example: string) {
   }
 
 function CodingEnvironment() {
+
+  /**
+  * Helper function to check if input contains only numbers.
+  */ 
+   function checkInputField(jsonField: JSON, key: string) {
+
+    const jsonInput = jsonField[key as keyof typeof jsonField];
+
+    if (!Array.isArray(jsonInput)) {
+
+      const errorMessage = `If you use ${key},
+      it must be an array of numbers.`;
+
+      toast.error(errorMessage);
+
+      setOutput(errorMessage);
+
+      return false;
+
+    }
+
+    if (Object.values(jsonInput).length === 0 || Object.values(jsonInput).some(isNaN)) {
+
+      const errorMessage = `If you use ${key}, 
+      it must contain at least one number, 
+      and it can only contain numbers.`;
+
+      toast.error(errorMessage);
+
+      setOutput(errorMessage);
+
+      return false;
+    }
+
+    return true;
+
+    }
+
+
+  /**
+  * We check the inputs and return true or false. We allow: 
+  * - an empty input, and
+  * - a valid JSON object containing stack_init or advice_tape (or both) 
+  * if the values are numbers
+  */ 
+  function checkInputs(jsonString: string) {
+
+    if (jsonString === "") {
+      return true;
+    }
+
+    let jsonInput!: JSON;
+
+    try {
+
+      jsonInput = JSON.parse(jsonString);
+
+    }
+    catch (e: any) {
+
+      const errorMessage = `Miden VM Inputs need to be a valid JSON object:
+${e.message}`;
+
+      toast.error(errorMessage);
+
+      setOutput(errorMessage);
+
+      return false;
+
+    }
+
+    if (!Object.keys(jsonInput).includes("stack_init") && !Object.keys(jsonInput).includes("advice_tape")) {
+
+      const errorMessage = `Miden VM Inputs can be empty or 
+we need either a stack_init or 
+an advice_tape.`;
+
+      toast.error(errorMessage);
+
+      setOutput(errorMessage);
+
+      return false;
+
+    }
+    Object.keys(jsonInput).forEach(key => {
+      checkInputField(jsonInput, key);
+    });
+
+    return true;
+  }
 
   /**
   * This sets the inputs to the default values.
@@ -76,13 +167,17 @@ end`
   const runProgram = async () => {
     init().then(() => {
       try {
-        const resp = run_program(code, inputs, numOfOutputs);
-        console.log(resp.toString());
-        setOutput(`Miden VM Program Output
+        if (checkInputs(inputs)) {
+
+          const resp = run_program(code, inputs, numOfOutputs);
+
+          setOutput(`Miden VM Program Output
 Stack = [${resp.toString()}]
 Cycles = <available with Miden VM v0.4>`);
+
+        }
       } catch (error) {
-        setOutput("Error: Check the developer console for details.");
+          setOutput("Error: Check the developer console for details.");
       }
         })};
   
@@ -92,26 +187,32 @@ Cycles = <available with Miden VM v0.4>`);
   * It runs the Rust program that is imported above.
   */ 
 
-    const proveProgram = async () => {
-      init().then(() => {
-        try {
+  const proveProgram = async () => {
+    init().then(() => {
+      try {
+        if (checkInputs(inputs)) {
+
           const resp = prove_program(code, inputs, numOfOutputs);
-          console.log(resp.toString());
           const stack_output = resp.slice(0, numOfOutputs);
           const overflow_addrs = resp.slice(numOfOutputs, resp.length);
+
           setOutput(`Miden VM Program Output
 Stack = [${stack_output.toString()}]
 Overflow Address = [${overflow_addrs.toString()}]
 Cycles = <available with Miden VM v0.4>`);
-        } catch (error) {
-          setOutput("Error: Check the developer console for details.");
+
         }
+      } catch (error) {
+        setOutput("Error: Check the developer console for details.");
+      }
           })};
     
   /**
-  * We need to add the following:
-  * Verify, and Debug as functions that can be called.
-  */ 
+  * We need to add logic to Verify, and Debug.
+  */   
+  const debugProgram = async () => {  init().then(() => { toast.error("Not yet available")})};
+  const verifyProgram = async () => {  init().then(() => { toast.error("Not yet available")})}; 
+
 
   return (
         <>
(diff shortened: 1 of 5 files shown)


Dominik1999:
Hmm added a new formatter. Only changes to `App.tsx` are relevant for that PR. The rest is formatting. 

I can change the PR only to contain the changes and then send a PR with the new formatting.

grjte:
> Hmm added a new formatter. Only changes to `App.tsx` are relevant for that PR. The rest is formatting.
> 
> I can change the PR only to contain the changes and then send a PR with the new formatting.

That would be great, thanks. It'll be easier to review & a cleaner git history to keep them separate

playground/src/App.tsx
```diff
@@ -96,26 +174,34 @@ Cycles = <available with Miden VM v0.4>`);
...
+      } catch (error) {
+        setOutput("Error: Check the developer console for details.");
+      }
+        })};
     
   /**
   * We need to add the following:
   * Verify, and Debug as functions that can be called.
-  */ 
+  */ ``
```
grjte, line 204: Why do we need these?

playground/src/App.tsx
```diff
@@ -96,26 +174,34 @@ Cycles = <available with Miden VM v0.4>`);
...
+        if (!tryParseJSONObject(inputs)) {
+          setOutput(`Error: Invalid JSON input`);
+          return;
+        }
+        if (!checkStackInitAndAdviceTape(inputs)) {
+          setOutput(`Error: Invalid inputs - no stack_init or advice_tape`);
+          return;
+        }
+        const resp = prove_program(code, inputs, numOfOutputs);
+        console.log(resp.toString());
```
grjte, line 189: We can probably remove this console log

playground/src/App.tsx
```diff
@@ -96,26 +174,34 @@ Cycles = <available with Miden VM v0.4>`);
...
+    init().then(() => {
+      try {
+        if (!tryParseJSONObject(inputs)) {
+          setOutput(`Error: Invalid JSON input`);
+          return;
+        }
+        if (!checkStackInitAndAdviceTape(inputs)) {
+          setOutput(`Error: Invalid inputs - no stack_init or advice_tape`);
+          return;
+        }
```
grjte, line 187: This code is duplicated above. Let's make a helper

playground/src/App.tsx
```diff
@@ -16,6 +17,75 @@ async function getExample(example: string) {
...
+  if (jsonString === "") { 
+    return true 
+  }
+
+  if (!jsonString.includes("stack_init") && !jsonString.includes("advice_tape")) {
+    toast.error('Miden VM Inputs can be empty or we need either a stack_init or an advice_tape.');
+    return false
+  }
+
+  const o = JSON.parse(jsonString);
```
grjte, line 66: There's some duplication between these. Is there a reason we can't combine them into a single `checkInputs` function?

grjte (changes requested):
Thanks @Dominik1999! I didn't try running it up locally yet, so this review is just based on looking at the code. Once you've addressed these comments I can take a quick look at that too

playground/src/App.tsx
Dominik1999: jup that works as well

playground/src/App.tsx
Dominik1999: jup

playground/src/App.tsx
Dominik1999: better?

Dominik1999:
> Thanks @Dominik1999! I didn't try running it up locally yet, so this review is just based on looking at the code. Once you've addressed these comments I can take a quick look at that too

Thanks for your review. Your points should be addressed by now. Let me know what you think. 

I tested it locally, but we need to add proper testing to the whole playground, which will be reflected in the next PR - issue here https://github.com/0xPolygonMiden/examples/issues/28

playground/src/App.tsx
```diff
@@ -16,6 +17,79 @@ async function getExample(example: string) {
...
+    toast.error('Miden VM Inputs can be empty or we need either a stack_init or an advice_tape.');
+    return false;
+
+  }
+
+  const o = tryParseJSONObject(jsonString);
+
+  if (o === false) {
+    return false;
+  }
```
grjte, line 68: It looks like this if statement wouldn't be necessary if the contents of `tryParseJSONObject` were moved in here. Given that it's not a long function and it's not used anywhere else (unless I missed it), I would probably prefer to move that completely into here.

I think a few other things could probably be condensed a bit here too. For example, we're doing the checks to see if the string includes "stack_init" and "advice_tape" twice.

grjte (changes requested):
Thanks! A few similar additional comments. Also, we shouldn't include the `.DS_Store` file

Overcastan (approved):
Looks good to me, thank you @Dominik1999! 
Just one small nit

playground/src/App.tsx
Dominik1999: Got it, refactored quite a bit.

playground/src/App.tsx
```diff
@@ -16,6 +17,79 @@ async function getExample(example: string) {
...
+
+  if (jsonString.includes("advice_tape")) {
+    if (o.advice_tape.len === 0 || o.advice_tape.some(isNaN)) {
+      toast.error(`If you use advice_tape, 
+      it must contain at least one number, 
+      and can only contain numbers.`);
+      return false;
+    }
+  }
```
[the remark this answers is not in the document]
Dominik1999: done

playground/src/App.tsx
```diff
@@ -16,6 +17,79 @@ async function getExample(example: string) {
...
+    return false;
+
+  }
+}
+
+/**
+* We check the inputs and return true or false. We allow: 
+* - an empty input, and
+* - a valid JSON object containing stack_init or advice_tape (or both) 
+* iff the values are numbers
```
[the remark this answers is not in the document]
hackaugusto: `iff` can also be the short hand for `if and only if` [ref](https://www.thefreedictionary.com/iff)

Dominik1999:
I will rebase

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -15,6 +16,98 @@ async function getExample(example: string) {
 
 function CodingEnvironment() {
 
+   /**
+  * Helper function to check if input contains only numbers.
+  */
```
hackaugusto, line 21: please fix the indentation. IMO we should have [prettier](https://prettier.io/docs/en/install.html) or something like it in the CI / pre-commit hooks.

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -15,6 +16,98 @@ async function getExample(example: string) {
...
+  * - a valid JSON object containing stack_init or advice_tape (or both) 
+  * if the values are numbers
+  */ 
+  function checkInputs(jsonString: string) {
+
+    if (jsonString === "") {
+      return true;
+    }
+
+    if (!jsonString.includes("stack_init") && !jsonString.includes("advice_tape")) {
```
hackaugusto, line 69: instead of checking to see if `stack_init` is a substring of `jsonString`. Instead it should check if `stack_init` is a key of the parsing result. (same for `advice_tape`)

hackaugusto (changes requested):
Please remove the `.DS_Store` file from the commit.

Also to prevent this, add it to the `.gitignore` file

playground/src/pages/CodingEnvironment.tsx
Dominik1999: Indeed. I already have it, but when I let it run now, the changes in this PR grow very big. 

I suggest I will run prettier after this merge in a second PR? But happy to do otherwise

hackaugusto: In that case I would just manually fix this and open the follow up later on

Dominik1999: got it

playground/src/pages/CodingEnvironment.tsx
Dominik1999: got it, is changed

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -15,6 +16,96 @@ async function getExample(example: string) {
...
+    let jsonInput!: JSON;
+
+    try {
+
+      jsonInput = JSON.parse(jsonString);
+
+    }
+    catch (e) {
+
+      const errorMessage = `Miden VM Inputs need to be
```
hackaugusto, line 78: This is working, but the error message is not great.

Could you please add the contents of `e`? It is a [`SyntaxError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError), `e.message` should be enough.

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -15,6 +16,96 @@ async function getExample(example: string) {
...
+
+      toast.error(errorMessage);
+
+      setOutput(errorMessage);
+
+      return false;
+
+    }
+
+    if (!Object.keys(jsonInput).includes("stack_init") && !Object.keys(jsonInput).includes("advice_tape")) {
```
hackaugusto, line 89: humm, this doesn't seem to be caugthing the case with only `advice_tape`:

```
{
    "advice_tape": ["0"]
}
```

The error I get is:

image

playground/src/pages/CodingEnvironment.tsx
Dominik1999: Isn't the errorMessage defined here the better approach? The SyntaxError is caused when the input is not a valid JSON, right?

playground/src/pages/CodingEnvironment.tsx
Dominik1999: damn good catch, I need to create tests for all cases. Thanks. 

I need to change that in the Rust file, we want to allow Inputs with only `advice_tape`

playground/src/pages/CodingEnvironment.tsx
Dominik1999: Got it, all good now.

I will add tests for all cases in #28 - which will be the next PR

hackaugusto (approved):
LGTM

PR #55, opened 2023-02-08
MASM algorithm for Catalan numbers using dynamic programming

jjcnn:
A MASM algorithm for computing the nth Catalan number, using dynamic programming.

See https://www.geeksforgeeks.org/program-nth-catalan-number/ for a definition


examples/catalan.masm
```diff
@@ -0,0 +1,160 @@
...
+        dup.1
+        # Need to use u32 division to truncate properly
+        # Safe to use unchecked, since k < i <= n+1, and n+1 has been asserted as u32
+        u32unchecked_div.2
+        # [ k/2, running_total, k, i, ... ]
+        # Initialize counter
+        push.0
+        # [ j, k/2, running_total, k, i, ... ]
+        # Calculate sum until penultimate summand
+        exec.check_continue_gt
```
frisitano, line 81: Something to consider here is that `gt` incurs a [high clock cycle overhead](https://0xpolygonmiden.github.io/miden-vm/user_docs/assembly/field_operations.html#comparison-operations) (18 cycles) compared to `eq` (1 cycle) /`neq` (2 cycles).  You may be able to refactor this check to use `neq`.  I believe by design j starts off as 0 which is lower than `k/2` and hence you can change your continue condition to check for inequality.  Once j == 2 then you know that !(j < k/2).   

```
# Check if the counter is less than the maximum
proc.check_continue_gt
    # [i, max]
    exec.dup1dup1
    gt
    # [ i < max , i, max ]
end
```

would be become

```
# Check if the counter is less than the maximum
proc.check_continue
    # [i, max]
    exec.dup1dup1
    neq
    # [ i < max , i, max ]
end
```

I implemented a similar pattern of using `gt` checks on my first miden script and was advised the same so I am just passing on the wisdom.

examples/catalan.masm
```diff
@@ -0,0 +1,160 @@
...
+        drop
+    end
+    # stack : [ C_i, i, ... ], mem : [ C_0, C_1, ..., C_(i-1) ]
+    swap
+    mem_store
+    # stack : [ ... ], mem : [ C_0, C_1, ..., C_i ]
+end
+
+# Check if the counter is less than or equal to the maximum
+proc.check_continue_gte
```
frisitano, line 130: Maybe you could refactor this check to not use `gte` for performance reasons as described above.

frisitano (approved):
This is a cool example!  I wasn't aware of Catalan numbers so thank you for bringing it to my attention.  I've checked this for correctness and it looks good.  I left a few minor performance comments inline.  It's up to you whether you want to make the changes but not obligatory seeing as it's just an example, however I don't imagine it would take you long if you have some free time.

examples/catalan.masm
jjcnn: It's a good catch either way. :-)

I've refactored both loop conditions to use `neq` which simplifies the code as a whole, at the cost of two additional instructions (`add.1` and `sub.1`) in the main function.

examples/catalan.masm
```diff
@@ -0,0 +1,151 @@
...
+#
+# example:
+#
+# input:
+# [9]
+#
+# output:
+# [4862]
+
+# Check if the counter is less than the maximum
```
hackaugusto, line 23: nit: we use `#!` to express documentation, example [from sys](https://github.com/0xPolygonMiden/miden-vm/blob/main/stdlib/asm/sys.masm#L1-L6)

examples/catalan.masm
```diff
@@ -0,0 +1,151 @@
...
+# Calculate C_i.
+# Assumes that all C_j for j < i have already been calculated and stored in memory
+proc.catalan_i
+    # stack : [ i, ... ], mem : [ C_0, C_1, ..., C_(i-1) ]
+    dup
+    push.0
+    eq
+    if.true
+        # Base case: C_0 = 1
+        # [ 0, ... ]
```
hackaugusto, line 56: comment: it seems you have the stack prior to the instruction is executed, in this case `i == 0` because of the test above, and the stack is `[0,...]` prior to `push.1`, later it will be `[1,0,...]`.

I opened a discussion on the miden-vm including this topic so we can get a style :) https://github.com/0xPolygonMiden/miden-vm/discussions/698

hackaugusto (commented):
This is awesome! I did a quick review and left some comments about style just for reference :)

examples/catalan.masm
jjcnn: Changed here: https://github.com/0xPolygonMiden/examples/pull/58. Thanks.

examples/catalan.masm
jjcnn: I prefer the style I've used, so I'll comment in the dicussion instead.

PR #58, opened 2023-02-14
Change inline comments to documentation comments for procedures

jjcnn:
Addressing @hackaugusto 's comments from here: https://github.com/0xPolygonMiden/examples/pull/55#discussion_r1105631646

### DIFF
--- a/examples/catalan.masm
+++ b/examples/catalan.masm
@@ -1,22 +1,22 @@
 # Computes the nth Catalan number using dynamic programming.
 # (See https://www.geeksforgeeks.org/program-nth-catalan-number/)
-# 
+#
 # Definition:
 #     C_0 = 1
 #     C_(n+1) = C_0 * C_n + C_1 * C_(n-1) + ... + C_n * C_0
 #             =  2 * C_0 * C_n + 2 * C_1 * C_(n-1) + ... + 2 * C_(n/2) * C_(n/2+1) (if n+1 is even)
 #             or 2 * C_0 * C_n + 2 * C_1 * C_(n-1) + ... + C_(n/2)^2               (if n+1 is odd)
-# 
+#
 # Input: The stack should contain a single element n.
 # Output: A stack with a single element which is the nth Catalan number C_n.
-# 
+#
 # C_i for 0 <= i <= n will be stored in memory at address i.
-# 
+#
 # example:
-# 
+#
 # input:
 # [9]
-# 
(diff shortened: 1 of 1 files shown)


hackaugusto (commented):
thanks! just one nit

jjcnn:
> thanks! just one nit

Yeah, but it felt wrong to leave it hanging. :-)

examples/catalan.masm
```diff
@@ -1,26 +1,26 @@
...
 # C_i for 0 <= i <= n will be stored in memory at address i.
-#
+# 
 # example:
-#
+# 
 # input:
 # [9]
-#
+#
```
hackaugusto, line 19: shouldn't we remove the whitespace at the end of the line?

hackaugusto (approved):
lgtm

examples/catalan.masm
```diff
@@ -1,26 +1,26 @@
...
+# 
 # example:
-#
+# 
 # input:
 # [9]
-#
+# 
 # output:
 # [4862]
```
bobbinth, line 21: Module-level comments can also start with `#!`. This way, they'll be included into auto-generated docs.

jjcnn: I tried that, but got the following error:

```
$ ../miden-vm/target/release/miden run -a examples/catalan.masm -n 1
============================================================
Run program
============================================================
Reading program file `examples/catalan.masm`
Compiling program... Failed to compile program - doc comments separated by line break
```

Then I read the docs, and trusted them enough to change the comments to regular ones. :-)

There's an empty line after the program comments. If I add `#!` to that line as well then it works, but presumably this will result in the entire program comment being treated as a procedure comment in the docs?

examples/catalan.masm
jjcnn: Fixed.

examples/catalan.masm
bobbinth: hmmm - that's a bug then. Let's file an issue.

bobbinth: Or actually, this functionality may be available only on the `next` branch.

jjcnn: You're right. My build was from `main` from 2 weeks ago. At the time I just used it to test my Rust installation, and didn't think about which branch I was on.

Comments have been fixed.

hackaugusto:
LGTM

PR #63, opened 2023-02-14
fix: let's fix the build

Dominik1999:


playground/package.json
```diff
@@ -6,14 +6,15 @@
   "dependencies": {
     "@codemirror/theme-one-dark": "^6.1.0",
     "@headlessui/react": "^1.7.7",
-    "@tailwindcss/typography": "^0.5.9",
     "@tailwindcss/forms": "^0.5.3",
+    "@tailwindcss/typography": "^0.5.9",
```
hackaugusto, line 10: nit: does this line need to move?

playground/src/App.tsx
```diff
@@ -3,6 +3,7 @@ import NavBar from "./components/NavBar";
 import CodingEnvironment from "./pages/CodingEnvironment";
 import InstructionTable from "./pages/InstructionTable";
 
+
```
hackaugusto, line 6: nit: remove this space

playground/webpack.config.js
```diff
@@ -39,4 +38,4 @@ module.exports = {
     filename: 'bundle.js',
     path: path.resolve(__dirname, 'dist'),
   },
-};
\ No newline at end of file
+};
```
hackaugusto, line 41: we should probably add `pre-commit` here too, lets see if that gets merged on `miden-vm` and then add here too.

hackaugusto (commented):
I ran it locally with:

```
git clone 'https://github.com/0xPolygonMiden/examples.git'
cd examples/playground
git checkout dominik_adding_a_sass_loader_for_katex
npm run build:miden
npm install
npm run start
```

and later

```
npm run build
npx webpack
```

the webpack build finished successfully and the dev version worked as intended.

just a few nits

playground/package.json
Dominik1999: What do you mean?

playground/webpack.config.js
Dominik1999: Got it, can we do that in a separate PR? This should just fix the build.

playground/package.json
hackaugusto: this dep was already set, but a line above. So the change was just moving the dependency one line down. This is not a big deal, but it makes it harder to navigate in the code's history.

hackaugusto (approved):
LGTM

playground/package.json
Dominik1999: Got it, good point. I was copying a lot of dependencies for this PR so I forgot about that.

PR #71, opened 2023-02-23
fix: fixing build again

Dominik1999:
gh-pages.yml did define the wrong order of actions. We first need to build miden-wasm and then install everything using npm

See https://github.com/0xPolygonMiden/examples/actions/runs/4254305275/jobs/7400575591


hackaugusto (approved):
LGTM

PR #73, opened 2023-03-01
update: Updating to Miden v0.4 and concurrent runs

Dominik1999:
I am updating to Miden v0.4 and changing it to include the breaking changes.

### DIFF
--- a/benchmarking-cli/src/main.rs
+++ b/benchmarking-cli/src/main.rs
@@ -1,6 +1,7 @@
 use clap::Parser;
-use miden::{Assembler, ProgramInputs, ProofOptions};
-use winter_math::StarkField;
+use miden_vm::{
+    AdviceInputs, Assembler, Kernel, MemAdviceProvider, ProgramInfo, ProofOptions, StackInputs,
+};
 use std::fs;
 use std::time::Instant;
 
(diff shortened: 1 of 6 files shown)


benchmarking-cli/src/main.rs
```diff
@@ -87,46 +86,52 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
     let now = Instant::now();
     let _program2 = assembler
         .compile(&program_string)
         .expect("Could not compile source");
 
     println! {"Compilation Time (hot): {} ms", now.elapsed().as_millis()}
 
+
+    let stack_input_cloned = stack_input.clone();
```
Dominik1999, line 120: I am unsure if I need to `clone()` those. In Miden v0.3 I was able to borrow but that seems not possible anymore.

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -91,10 +91,12 @@ an advice_tape.`;
 
       return false;
     }
+
     Object.keys(jsonInput).forEach((key) => {
```
Dominik1999, line 95: Added this error fix - which has nothing to do with the Updage to Miden v0.4. Just saw it.

benchmarking-cli/src/main.rs
hackaugusto: It seems the API changed and now a clone is required.

hackaugusto (approved):
LGTM

PR #74, opened 2023-03-02
feat: Sending structs from Rust to TS

### LINKED ISSUE #52: Send proof from backend to frontend
## Goal
Currently, in the playground (after we merge #50) we can prove a program. We only return a simple `Vec<u64>` in which we encode the `stack_output` and the `overflow_addresses` to the front end to show it to the user. 

But we also want to return 
- the proof 
- the program hash
- metrics (cycles and time used to prove)

## Problem wasm-bindgen doesn't allow to return tuples
We can't simply create a tuple that includes all those values (including the proof in bytes). Wasm-bindgen (which compiles Rust to WASM) doesn't allow tuples as returns, see [here](https://stackoverflow.com/questions/75064800/how-to-return-a-tuple-to-frontend-from-wasm)

## Possible solution
There is a way to return a struct. So we might even be able to return `ProgramOutputs` (for Miden VM v0.3.) as a struct.  Here are the docs (https://rustwasm.github.io/wasm-bindgen/contributing/design/exporting-rust-struct.html). I only spent 1h, but I couldn't figure out how to implement this yet. 

## Further considerations 
We also need to find a way to show the proof to the user. Together with some links or explainers of what a proof is. Ideally, if a user would change the proof somehow, he could test to verify the changed proof and see that it would get rejected.

Dominik1999:
Closes #52

I created a new `struct` in the Rust backend called `Outputs`. Outputs are returned to the React frontend when running `run_program` or `prove_program`. Outputs include

```Rust
#[wasm_bindgen(getter_with_clone)]
pub struct Outputs {
    pub stack_output: Vec<u64>, 
    pub trace_length: usize,
    pub program_info: Option<Vec<u8>>,
    pub overflow_addrs: Option<Vec<u64>>,
    pub proof: Option<Vec<u8>>,
}
``` 

The proof is stored in the browser for later use 

```Typescript
        // Store the proof in the session storage
        if (proof) {
          sessionStorage.setItem("proof", proof.toString())  
          }
```


Overcastan (approved):
Looks good to me, thank you @Dominik1999!

Overcastan:
I think we should update the `Outputs` struct in the PR comment, since now it also stores `program_info`. But still looks good to me!

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -183,14 +183,22 @@ Cycles = <available with Miden VM v0.4>`);
...
+
+  // Store program_info in the session storage to verify the proof 
+          if (program_info) {
+            sessionStorage.setItem("program_info", program_info.toString())  
+            }      
+
+  // Store the proof in the session storage
+        if (proof) {
+          sessionStorage.setItem("proof", proof.toString())  
+          }
```
grjte, line 201: nit: the indentation here is a bit strange to look at

grjte (approved):
Looks good to me! A few things I noticed when running it locally:

- In the outputs, it says "Overflow Address". Instead, it should say "Overflow addresses"
- When the playground is processing after a button has been pressed, we should indicate somehow that things are processing, either with a spinner on the pressed button or a spinner in front of the whole page. All buttons should also be disabled until the playground is done processing.

PR #76, opened 2023-03-06
feat: Verification of programs

### LINKED ISSUE #67: Add Verify to the Miden Playground
We want to be able to verify programs in the Miden Playground. And we want that users are able to tamper with the Output that gets verified to see that not every program can be verified.

Therefore, we need to 
- recollect the values from the Output `stack_output` and `overflow_addr` from the frontend
- recollect the program inputs from the frontend
- pass the proof back to the backend, (need to do https://github.com/0xPolygonMiden/examples/issues/52 first)
- pass the program back to the backend to get the program.hash

It should fail if a user changes the input or output in the CodeMirror components.

Dominik1999:
I added the ability to verify a program to the playground. This should close #67 

The user can create a proof using the "Prove" button. Then the user can then verify that proof using the "Verify" button. 
If the Inputs, the Program or the Outputs are being tampered with after the proof was generated, the proof fails. 

https://user-images.githubusercontent.com/35031754/223128878-e07cb7d3-0d44-445f-a22a-82f5dcba9c8b.mov

**Note:** I needed to return the whole stack, for now, to be able to use the program outputs as inputs in the verification step. This is a bad UX.

```
  /**
   * This handles the number of outputs that are returned to the user.
   * At the moment we return the whole stack always. This is not good UX.
   * But it is needed to use the Output again as Input for the verification step.
   */
  // const [numOutputs, setNumOutputs] = React.useState(16);
```


playground/src/pages/CodingEnvironment.tsx
```diff
@@ -122,16 +122,15 @@ end`;
...
-
   const emptyOutput = "\n \n \n \n";
   const [output, setOutput] = React.useState(emptyOutput);
 
   /**
    * This handles the number of outputs that are returned to the user.
-   * It is not optimal to fix that to a certain number, but for
-   * the sake of simplicity we do it for now.
+   * At the moment we return the whole stack always. This is not good UX.
+   * But it is needed to use the Output again as Input for the verification step.
```
grjte, line 131: I'm not sure that there's a problem with returning the entire stack. As you said, we need it because the stack output and the overflow addresses are inputs for verification, and we want people to be able to understand how this is working. Maybe the UX improvement would be to show a section for "verifier inputs", which would contain the full stack + overflow_addrs, etc. Then we could still show a result as just the top # of stack outputs. I think we'd also want to add some hover info messages to explain everything for clarity.

I would probably remove this comment but make an issue for thinking about this

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -122,16 +122,15 @@ end`;
...
    */
-  const numOfOutputs = 16;
+  // const [numOutputs, setNumOutputs] = React.useState(16);
```
grjte, line 133: it looks like we can remove this

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -214,9 +212,33 @@ Cycles = [${trace_length.toString()}]`);
...
+          setOutput(`
+Verification success with security level of ${result}
+
+`);
+        }
+
+        toast.success("Verification successful");
+
+      } catch (error) {
+      setOutput("Error: Check the developer console for details.");
```
grjte, line 237: This can be updated in a separate PR, but I think we can stop forcing people to use the console and start showing the actual error now that we have toast messages and better error info being passed from `miden-wasm`. Same thing in `runProgram` and `proveProgram`

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -214,9 +212,33 @@ Cycles = [${trace_length.toString()}]`);
...
+   */
   const verifyProgram = async () => {
     init().then(() => {
-      toast.error("Not yet available");
+      try {
+
+        if (checkInputs(inputs)) {
+          const result = verify_program(code, inputs, output, proof);
+          setOutput(`
+Verification success with security level of ${result}
```
grjte, line 229: nit: We should talk about the bits of security rather than just showing a number. I would change this to "Verification succeeded with a security level of ${result} bits."

Note for the future: we should be able to easily make it so people can change the security level. I think this would be cool to do.

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -214,9 +212,33 @@ Cycles = [${trace_length.toString()}]`);
...
+   * It returns true if the proof is valid, and false otherwise.
+   * As inputs we need program_info, stack_input, stack_output, and proof.
+   */
   const verifyProgram = async () => {
     init().then(() => {
-      toast.error("Not yet available");
+      try {
+
+        if (checkInputs(inputs)) {
+          const result = verify_program(code, inputs, output, proof);
```
grjte, line 227: it looks like you're assuming that both the output and the proof are available. We should check for these just like we check inputs, and then raise a toast message if they're missing telling people to prove a program first

playground/miden-wasm/src/lib.rs
```diff
@@ -1,9 +1,9 @@
 mod expected_test_proof;
 use miden_vm::math::{Felt, FieldElement};
+use miden_vm::utils::Serializable;
```
grjte, line 3: I'm getting an unused import warning for this line

grjte (changes requested):
Thanks @Dominik1999 it's really cool to have this working! I left a few comments inline. Additionally, this isn't running successfully for me locally when I check out your branch. Can you double-check installing & building everything from scratch to see if you get errors?

playground/src/pages/CodingEnvironment.tsx
Dominik1999: Indeed, let's get back to it when we talk in-depth about UX

playground/src/pages/CodingEnvironment.tsx
Dominik1999: I think I now always set the Output to the corresponding error message.

playground/src/pages/CodingEnvironment.tsx
Dominik1999: very nice idea!

playground/src/pages/CodingEnvironment.tsx
Dominik1999: did it

Dominik1999:
> Thanks @Dominik1999 it's really cool to have this working! I left a few comments inline. Additionally, this isn't running successfully for me locally when I check out your branch. Can you double-check installing & building everything from scratch to see if you get errors?

Thanks a lot @grjte! 

I refactored a bit and addressed your comments. When I run `npm run build:miden && npm install && npm run start` in the playground folder it builds and starts. Which error message are you seeing?

playground/__tests__/Components.test.tsx
```diff
@@ -1,7 +1,13 @@
-import { fireEvent, render, screen, cleanup, within } from "@testing-library/react";
-import '@testing-library/jest-dom'
+import {
```
Dominik1999, line 1: This is only made prettier using `npx prettier --write .`. I wasn't able to undo that step. I can manually remove those files from the PR if you want.

playground/jest.config.ts
```diff
@@ -1,7 +1,7 @@
 module.exports = {
```
Dominik1999, line 1: This is only made prettier using `npx prettier --write .`. I wasn't able to undo that step. I can manually remove those files from the PR if you want.

playground/src/components/DropDown.tsx
```diff
@@ -18,11 +18,13 @@ type DropDownProps = {
   onExampleValueChange?: (newType: string) => void;
 };
 
-export default function DropDown({ onExampleValueChange }: DropDownProps): JSX.Element {
+export default function DropDown({
```
Dominik1999, line 21: This is only made prettier using `npx prettier --write .`. I wasn't able to undo that step. I can manually remove those files from the PR if you want.

playground/src/components/NavBar.tsx
```diff
@@ -6,7 +6,10 @@ const NavBar = () => {
   return (
     <>
       <div className="flex items-center justify-between flex-wrap bg-blue-700 p-6">
-        <div className="flex items-center flex-shrink-0 text-white mr-6" data-testid="logo">
+        <div
```
Dominik1999, line 9: This is only made prettier using `npx prettier --write .`. I wasn't able to undo that step. I can manually remove those files from the PR if you want.

playground/tsconfig.json
```diff
@@ -15,8 +15,7 @@
     "isolatedModules": true,
     "noEmit": true,
     "jsx": "react-jsx",
-    "types": ["node", "jest", "@testing-library/jest-dom"],
+    "types": ["node", "jest", "@testing-library/jest-dom"]
```
Dominik1999, line 18: This is only made prettier using `npx prettier --write .`. I wasn't able to undo that step. I can manually remove those files from the PR if you want.

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -1,109 +1,134 @@
...
+   * - an empty input, and
+   * - a valid JSON object containing stack_init or advice_tape (or both)
+   * if the values are numbers
    */
-  function checkInputField(jsonField: JSON, key: string) {
-    const jsonInput = jsonField[key as keyof typeof jsonField];
+  function checkInputs(jsonString: string) {
+    if (jsonString === "") {
+      return true;
+    }
```
grjte, line 26: This function and `checkOutputs` are both quite long and I think it would be better to put them into the utils. This could be done if just used toast messages instead of putting the error message in the output. Personally, I think this might be fine. These are checks that run when the user makes a request and they aren't output errors like invalid proofs, execution failure, etc. Alternatively, you could pass `setOutput` into the function so that these could still be moved to utils without changing the current UX

So to summarize - I would prefer to move both `checkInputs` and `checkOutputs` to utils

grjte (changes requested):

playground/src/pages/CodingEnvironment.tsx
Dominik1999: Jep, I think you are right. Let's do that in the next refactoring PR if you don't mind. This PR is already quite big. 

What do you think?

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -214,9 +232,25 @@ Cycles = [${trace_length.toString()}]`);
...
     init().then(() => {
-      toast.error("Not yet available");
+      try {
+        if (checkInputs(inputs) && checkOutputs(outputs, proof)) {
+          const result = verify_program(code, inputs, outputs, proof);
+          setOutput(`Verification succeeded with a security level of ${result} bits. \n \n \n`);
+          toast.success("Verification successful");
+        }
+      } catch (error) {
+          setOutput("Error: Check the developer console for details.");
```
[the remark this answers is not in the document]
Dominik1999: Understand that point. I think we don't get the error here from Rust atm. If the verification fails, we get `unreachable`. There might be a way to propagate the error as well. I can include that into the PR or we can as well work on that in the next UX PR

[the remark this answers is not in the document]
grjte: Just so I understand - the error in the console is coming from the wasm package and you don't have access to it? If that's correct, then yes, we could tackle that as its own issue. I don't think we should have a single "UX PR". I think we'll want to break it down into a bunch of small steps, and this could be one of them

playground/src/pages/CodingEnvironment.tsx
grjte: Ok

playground/src/pages/CodingEnvironment.tsx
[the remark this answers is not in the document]
Dominik1999: You are correct. I don't have access to the wasm error (yet). I am sure somehow we can propagate it. 

And yes, we should have many small UX PRs

grjte (approved):
As discussed, we'll address the final issues separately in future refactor and UX PRs, so I'm approving this

PR #77, opened 2023-03-08
refactor: checkInput and checkOutputs to utils

Dominik1999:
As discussed in https://github.com/0xPolygonMiden/examples/pull/76#discussion_r1129264774

### DIFF
--- a/playground/src/pages/CodingEnvironment.tsx
+++ b/playground/src/pages/CodingEnvironment.tsx
@@ -11,121 +11,13 @@ import init, {
   verify_program,
 } from "miden-wasm";
 import toast, { Toaster } from "react-hot-toast";
-import { getExample, checkFields } from "../utils/helper_functions";
+import {
+  getExample,
+  checkInputs,
+  checkOutputs,
+} from "../utils/helper_functions";
 
 function CodingEnvironment(): JSX.Element {
-  /**
-   * We check the inputs and return true or false. We allow:
-   * - an empty input, and
-   * - a valid JSON object containing stack_init or advice_tape (or both)
-   * if the values are numbers
-   */
-  function checkInputs(jsonString: string) {
-    if (jsonString === "") {
-      return true;
-    }
-
-    let jsonInput!: JSON;
-
-    try {
-      jsonInput = JSON.parse(jsonString);
-    } catch (e: any) {// eslint-disable-line @typescript-eslint/no-explicit-any
-      const errorMessage = `Miden VM Inputs need to be a valid JSON object:
-${e.message}`;
-
-      toast.error(errorMessage);
-
-      setOutput(errorMessage);
-
-      return false;
-    }
-
-    if (
-      !Object.keys(jsonInput).includes("stack_init") &&
-      !Object.keys(jsonInput).includes("advice_tape")
-    ) {
-      const errorMessage = `Miden VM Inputs can be empty or 
-we need either a stack_init or 
-an advice_tape.`;
-
-      toast.error(errorMessage);
-
-      setOutput(errorMessage);
-
-      return false;
-    }
-
-    if (!checkFields(jsonInput)[0]) {
-      const errorMessage = checkFields(jsonInput)[1]
-      toast.error(errorMessage);
-      setOutput(errorMessage);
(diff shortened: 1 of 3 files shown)


playground/src/pages/CodingEnvironment.tsx
```diff
@@ -242,10 +134,18 @@ end`;
...
           setOutput(`Verification succeeded with a security level of ${result} bits. \n \n \n`);
           toast.success("Verification successful");
+        } else {
+            if (!checkInputs(inputs)[0]) { 
+              setOutput(checkInputs(inputs)[1]);
+              toast.error("Verification failed"); 
+            } else { 
+                setOutput(checkOutputs(outputs, proof)[1]);
+                toast.error("Verification failed"); 
+              }
```
grjte, line 148: Instead of indexing like this, I think it would be clearer to define and return an object type with a boolean `isValid` and a string "error". Also`checkInputs` is being run three times in the error case. I would just run it once and get a result. Finally, you can condense things a bit with an "else if"

Something like the following:

```
const inputCheck = checkInputs(inputs);
const outputCheck = checkOutputs(outputs, proof);
if (inputCheck.isValid && outputCheck.isValid) {
          const result = verify_program(code, inputs, outputs, proof);
          setOutput(`Verification succeeded with a security level of ${result} bits. \n \n \n`);
          toast.success("Verification successful");
} else if (!inputCheck.isValid) { 
              setOutput(inputCheck.error);
              toast.error("Verification failed"); 
} else { 
                setOutput(outputCheck.error);
                toast.error("Verification failed"); 
}
```

Also some of the indentation seems kind of wonky. Shouldn't this be handled by prettier now?

grjte (changes requested):
Thanks @Dominik1999, I left a proposed change inline

playground/src/pages/CodingEnvironment.tsx
Dominik1999: thanks that is a good suggestion

Dominik1999: If the success case is unnested, would it always be executed?

Dominik1999: I changed it and return early

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -176,16 +68,19 @@ end`;
 
   const runProgram = async () => {
     init().then(() => {
+      const inputCheck = checkInputs(inputs);
+        if (!inputCheck.isValid) {
+          setOutput(inputCheck.errorMessage);
+          toast.error("Running failed");
```
grjte, line 74: tiny nit: I would probably say "Execution failed"

playground/src/utils/helper_functions.tsx
```diff
@@ -55,8 +59,101 @@ export function checkField(jsonField: JSON, key: string): [boolean, string] {
...
+an advice_tape.`;
+
+    return { isValid: false, errorMessage: errorMessage };
+  }
+
+  if (!checkFields(jsonInput).isValid) {
+    const errorMessage = checkFields(jsonInput).errorMessage;
+
+    return { isValid: false, errorMessage: errorMessage };
+  }
```
grjte, line 105: In general, I would avoid calling the same function multiple times. Instead, you can call it once, get the value, then check that value and already have the `errorMessage` available if you needed. Then you can also just return that value instead of constructing the return object. This is similar to the change from the last iteration, and I would make the same change in the `checkOutputs` function below.

grjte (changes requested):
Thanks @Dominik1999, I think this is cleaner & looks good. I left one more similar comment inline and an optional nit

playground/src/pages/CodingEnvironment.tsx
Dominik1999: that is better than "Running" indeed

playground/src/utils/helper_functions.tsx
Dominik1999: sounds good

grjte (approved):
Looks good to me now, thanks!

PR #79, opened 2023-03-13
feat: adds debugging with no UI [WIP]

### LINKED ISSUE #68: Add Debug for the Miden Assembly Playground
We need to add a function to Debug to the playground. The Button is already there. 

We can use the simple assembler debugger here https://github.com/0xPolygonMiden/miden-vm/pull/693.

However, we need to wrap it into a function to call something like

```Rust
     pub fn debug_program(masm_code: &str, inputs_frontend: &str, command: &str) -> Vec<u64> {
     ...
    }
```

Where command is `!next`, `!play`, `!prev`, `!print`.

We will not be able to support all commands at the beginning. 
_____

We can simply call `execute_iter()` in here
https://github.com/0xPolygonMiden/miden-vm/blob/360b5b080b79b9d66e912e2eb286a5237a261eb2/processor/src/lib.rs#L118-L138

Then we have a `VmStateIterator` object on which we can call various debugging commands, see https://github.com/0xPolygonMiden/miden-vm/blob/360b5b080b79b9d66e912e2eb286a5237a261eb2/processor/src/debug.rs#L128

The challenging question will be how the `VmStateIterator` can be stored in the state of the browser.

Dominik1999:
closes #68 

This PR enables the frontend to run the Rust program `debug_program` which executes `miden_vm::execute_iter`.

Because the `VmStateIterator` can not yet be passed to the frontend, the frontend stores the number of steps already executed and passes this as a variable to the Rust program. That means when the user clicks the first time, it executes step 1 and displays the output. When the user clicks the second time, it executes step 1 and step 2 and displays the output.

This should not be the final version. However, in order to have smaller PRs, I want to merge the functionality and, in a second  PR the corresponding UX changes.


playground/miden-wasm/src/lib.rs
```diff
@@ -142,6 +159,67 @@ pub fn prove_program(asm: &str, inputs_frontend: &str) -> Outputs {
     result
 }
 
+/// Debuggs the program with the given inputs
+#[wasm_bindgen]
+pub fn debug_program(asm: &str, inputs_frontend: &str, steps: usize) -> DebugOutputs {
```
hackaugusto, line 164: This seems quite inefficient, for example, lets say we are on step 10_000, to get into step 10_001 it seems the program has to be instantiated again and re-executed up to the target step.

Can't we keep a reference of the previous program in memory and just advance a single step?

Dominik1999: Indeed. That's the problem. We need to create a `VmStateIterator` in the front end. Over the weekend, I could not find out how to do that. Especially because we need all the types of that object replicated in Typescript.

Dominik1999: Ok, let me try. But I am not sure if we can export this. At least in the Game of Life example, the `Universe` structure only consists of `String` and `i32` types, whereas the `VmStateIterator` consists of many types that I am not allowed to export to JavaScript land.

bitwalker: @Dominik1999 An approach you can use is rather than create a binding directly for `VmStateIterator`, create a handle type which carries a pointer to the `VmStateIterator` (typically done by boxing the value, but you can use references/pointers too, whatever is most appropriate). This can be represented trivially in the Wasm bindings, and then all you need to do is expose functions on that handle type to drive the iterator as normal. It's a layer of indirection, but it should be easier to generate bindings for.

playground/miden-wasm/src/expected_test_proof.rs
```diff
@@ -1,3 +1,2599 @@
 // quite a long proof, but it's just a test
```
Dominik1999, line 1: I ran `cargo fmt` and that got updated. I can check that file out of the commit if you want.

bitwalker (approved):
No notes, LGTM!

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -38,13 +42,24 @@ function CodingEnvironment(): JSX.Element {
    */
   const [proof, setProof] = useState<Uint8Array>(new Uint8Array(0));
 
+  /**
+   * This helps us to show the Debug Menu.
```
hackaugusto, line 46: nit: it is better to now use the first person on documentation. Suggestion:

> Determines when to show the debug menu

(ref from the rust style guidelines, I believe this should apply to any language though)

https://doc.rust-lang.org/1.0.0/style/style/comments.html#sentence-structure

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -38,13 +42,24 @@ function CodingEnvironment(): JSX.Element {
...
+  const [debugExecutor, setDebugExecutor] = useState<DebugExecutor | null >(null);
+
   /**
    * This handles a change in the selected example.
    * If a new example is selected using the dropdown, the inputs and 
    * the code are updated.
    */
   const [, setExample] = React.useState<string>();
   const handleSelectChange = async (exampleChange: string) => {
+    setShowDebug(false);
```
hackaugusto, line 62: I think it makes sense to remove the executor state too. perhaps moves this to an utility function, say, `disable_debug`, that would also call `setDebugExecutor(null)`?

playground/src/utils/helper_functions.tsx
```diff
@@ -145,3 +145,24 @@ Did you prove the program first?`;
 
   return checkFields(jsonOutput)
 }
+
+/**
+ * Helper function to format the debug output.
+ */
+export function addNewlineAfterWhitespace(str: string): string {
```
hackaugusto, line 152: question:

- why not return json from the rust side so that js can format as it wants?
- or to format everything as it should in the rust side so that it is in a single place?

playground/miden-wasm/src/lib.rs
```diff
@@ -142,22 +142,226 @@ pub fn prove_program(asm: &str, inputs_frontend: &str) -> Outputs {
...
+        return format!(
+            "{}",
+            self.vm_state
+                .stack
+                .iter()
+                .enumerate()
+                .map(|(i, f)| format!("[{i}] {f}"))
+                .collect::<Vec<_>>()
+                .join("\n"),
+        );
```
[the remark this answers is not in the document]
Dominik1999: This change results in 

```
error[E0308]: mismatched types
   --> src/lib.rs:299:38
    |
299 |               |s, (i, f)| s.push_str(format!("[{i}] {f}\n"))
    |                                      ^^^^^^^^^^^^^^^^^^^^^^ expected `&str`, found struct `String`
    |
    = note: this error originates in the macro `format` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
   --> src/lib.rs:299:27
    |
299 |               |s, (i, f)| s.push_str(format!("[{i}] {f}\n"))
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `String`, found `()`
    |
note: method `push_str` modifies its receiver in-place
   --> src/lib.rs:299:29
    |
299 |               |s, (i, f)| s.push_str(format!("[{i}] {f}\n"))
    |                           - ^^^^^^^^ this call modifies `s` in-place
    |                           |
    |                           you probably want to use this value after calling the method...
    = note: ...instead of the `()` output of method `push_str`
    ```

File: playground/miden-wasm/src/lib.rs [hunk 7]
```diff
@@ -142,22 +142,226 @@ pub fn prove_program(asm: &str, inputs_frontend: &str) -> Outputs {
...
+    }
+
+    // MODIFIERS
+    // --------------------------------------------------------------------------------------------
+
+    /// executes a debug command against the vm in it's current state.
+    pub fn execute(&mut self, command: i16, param: u64) -> String {
+        match command {
+            //DebugCommand::PlayAll
+            1 => {
```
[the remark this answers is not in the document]
Dominik1999: It looks like they are working on it but the feature is not there yet

https://github.com/rustwasm/wasm-bindgen/issues/2407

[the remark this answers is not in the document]
Dominik1999: we could do strings instead of numbers?

playground/src/utils/helper_functions.tsx
Dominik1999: I have added an issue for that. https://github.com/0xPolygonMiden/examples/issues/89

We should have this consistent with the output of the Run and Proof step. We can always return a JSON from Rust Land. 

I would not do that now because it is unclear which debugging outputs we need exactly. Once we have settled there, I will change it. Agreed?

playground/miden-wasm/src/lib.rs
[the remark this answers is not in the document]
hackaugusto: @Dominik1999 we don't need the full ADT version, but the C-style enum, which should be supported

[example test](https://github.com/rustwasm/wasm-bindgen/blob/main/tests/wasm/enums.rs#L15-L21), [example docs](https://rustwasm.github.io/docs/wasm-bindgen/reference/types/imported-js-types.html#example-rust-usage)

playground/miden-wasm/src/lib.rs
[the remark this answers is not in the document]
hackaugusto: oh, sorry, I didn't test the code, have you tried something like this?

```rust
|s, (i, f)| {
    s.push_str(&format!("[{i}] {f}\n"));
    s
}
```

playground/src/utils/helper_functions.tsx
hackaugusto: sounds good 👍

hackaugusto (approved):
LGTM, remaining comments are nits and can be done on follow ups. I ran it locally and tested with the `comparison` example program

PR #93, opened 2023-03-16
feat: CodeMirror fixed lines and more features

### LINKED ISSUE #85: CodeMirror component should always show the same number of lines by default
The CodeMirror Component should always have a certain number of lines by default, regardless of how big the code is. Then after a certain max_height, it should be scrollable. 

Screenshot 2023-03-14 at 19 46 38

In this example the Inputs CodeMirror box has 4 lines and the Outputs box only 1.

Dominik1999:
Closes #85 

Fixes the height of the code mirror to a certain number. 
Adds some features to the Code Window of the code mirror. Not all work yet - eg syntax highlighting - only when we define the Miden Assembly Syntax.

```
            basicSetup={{
              foldGutter: true,
              highlightActiveLineGutter: true,
              dropCursor: true,
              allowMultipleSelections: false,
              indentOnInput: false,
              lineNumbers: true,
              syntaxHighlighting: true,
              bracketMatching: true,
              autocompletion: true,
              highlightActiveLine: true,
            }}
```

It is weird but true that you cannot set the code mirror to a specific number of lines. This seems to be the workaround, see [here](https://discuss.codemirror.net/t/minimum-number-of-lines-dynamic-min-height/946/3).


hackaugusto:
@Dominik1999 could you explain a bit why this is necessary? what is being achieved by this change?

hackaugusto:
something is off, is this really necessary? When I load the playground I get this:

image

The inputs does not have the 8 lines, and has the same height as the outputs. Am I missing something?

hackaugusto:
question: do we want the two blocks to have the same line number, or the same height? These are different things, and we can achieve the later with this patch:

```patch
diff --git a/playground/src/pages/CodingEnvironment.tsx b/playground/src/pages/CodingEnvironment.tsx
index ef8e3e9..8e3ae13 100644
--- a/playground/src/pages/CodingEnvironment.tsx
+++ b/playground/src/pages/CodingEnvironment.tsx
@@ -261,6 +261,7 @@ export default function CodingEnvironment(): JSX.Element {
             <CodeMirror
               value={inputs}
               height="100%"
+              minHeight="150px"
               maxHeight="150px"
               theme={oneDark}
               onChange={setInputs}
@@ -271,8 +272,9 @@ export default function CodingEnvironment(): JSX.Element {
             <CodeMirror
               value={outputs}
               height="100%"
+              minHeight="150px"
               maxHeight="150px"
-              theme={eclipse}
+              theme={oneDark}
               onChange={setOutput}
             />
           </div>
```

It looks like this:

image

And it is a much cleaner fix

Dominik1999:
I want both to be honest, the same height and line number. And I want a third property that the CodeMirror shows 8 line numbers at the beginning. 

Screenshot 2023-03-17 at 17 31 38

Dominik1999:
When I run this commit locally I get the abovementioned picture.

hackaugusto (approved):
LGTM

Issue #86, opened 2023-03-14
Improve the Instruction Set table


Dominik1999:
There are a couple of improvements that we can do to the Instruction Set. 

- [x] We could remove `cycle` and `cycles` in the Cycles column and only have a number
- [x] We should be able that the instructions set pops out and overlays the CodingEnvironment. That way users could looks at the instructions and still code
- [x] We should add a search functionality to the table
- [x] We should update the table, some instructions are not correct, e.g. `rphash`
- [x] We should have the font in the table black

PR #100, opened 2023-03-17
Improve the Instruction Set table

Dominik1999:
Closes the easy parts of #86 

- added a search to the instructions table. 
- removed `cycles` from the Cycles column (its obvious)
- updated the instruction set
- changed the font in the table to black

### DIFF
--- a/playground/src/data/instructions.tsx
+++ b/playground/src/data/instructions.tsx
@@ -6,7 +6,7 @@ export const assemblerInstructions = [
         instruction: "assert",
         stackInput: "[a, ...]",
         stackOutput: "[...]",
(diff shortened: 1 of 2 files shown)


hackaugusto:
I think removing `cycles` word is not a good idea, this is how it looks on my laptop:

image

without the header the numbers are missing context.

Dominik1999:
Weird, this is how it renders on my computer


https://user-images.githubusercontent.com/35031754/226367287-618645aa-75af-4011-954b-791b6eb8a63e.mov

hackaugusto (approved):
LGTM, thanks!

PR #105, opened 2023-03-21
fix absolute URL 404 error

prabal-banerjee:
Change BrowserRouter to HashRouter as gh pages fail to resolve URLs otherwise

### DIFF
--- a/playground/package.json
+++ b/playground/package.json
@@ -1,7 +1,7 @@
 {
   "name": "miden-assembly-examples-playground",
   "version": "0.1.0",
-  "homepage": "https://0xpolygonmiden.github.io/examples/",
+  "homepage": "https://0xpolygonmiden.github.io/examples/#",
   "private": true,
   "dependencies": {
     "@codemirror/theme-one-dark": "^6.1.0",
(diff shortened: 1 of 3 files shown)


playground/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "miden-assembly-examples-playground",
   "version": "0.1.0",
-  "homepage": "https://0xpolygonmiden.github.io/examples/",
+  "homepage": "https://0xpolygonmiden.github.io/#",
```
Dominik1999, line 4: I think here you need to change to 

```
"homepage": "https://0xpolygonmiden.github.io/examples",
```

playground/src/App.tsx
```diff
@@ -7,19 +7,19 @@ import ExplainerPage from "./pages/Explainer";
 function App() {
   return (
     <>
-      <Router>
+      <HashRouter>
         <NavBar />
         <Routes>
           <Route path="/examples/" element={<CodingEnvironment />} />
```
Dominik1999, line 13: I have `path="/examples"` with the `/` at the end. Not sure if this is important?

Dominik1999 (changes requested):

Dominik1999:
@prabal-banerjee thanks for the nice fix! This helps a lot.

I think you need to keep the `homepage` in `package.json` to not break things.

Dominik1999 (approved):
looks good to me, thanks

Issue #95, opened 2023-03-17
Add toast.success when Executing and Proving


Dominik1999:
We can add a toast.success also after executing and proving. The toast can show the time it took.

Issue #75, opened 2023-03-06
Show the user that the playground is working - e.g. with a spinner


Dominik1999:
When the playground is processing after a button has been pressed, we should indicate somehow that things are processing, either with a spinner on the pressed button or a spinner in front of the whole page. All buttons should also be disabled until the playground is done processing.

PR #107, opened 2023-03-21
Dominik show proof and proving to user

### LINKED ISSUE #96: We can show to the user that a proof was generated
Because we don't show the proof to the user, atm the difference between executing and probing is quite subtle (or not visible in the example of Fibonacci). 

We can have a small icon that indicates that there is a proof. We can show the proof size and the security level of the proof.

In the future, we can provide the ability to download and even upload a proof.

The user should be able to see the proof in Hex.

Dominik1999:
Closes #96 and #95 and #75 

- We show to the use a spinner when its proving
- The proof can be shown to the user when its there with a new button (the proof is too long though - not the best UX)
- When the proof was generated, a symbol changes from red to green


playground/src/pages/CodingEnvironment.tsx
```diff
@@ -87,18 +89,21 @@ export default function CodingEnvironment(): JSX.Element {
...
         return;
       }
       try {
+        const start = Date.now();
         const { stack_output, trace_len }: Outputs = run_program(code, inputs);
         setOutput(`{
 "stack_output" : [${stack_output.toString()}],
 "trace_len" : ${trace_len}
 }`);
+      toast.success(`Execution successful in ${Date.now() - start} ms`);
```
hackaugusto, line 106: this is oddly indented, please double check you're applying formatting correctly

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -110,18 +115,21 @@ export default function CodingEnvironment(): JSX.Element {
    * It runs the Rust program that is imported above.
    */
   const proveProgram = async () => {
+    toast.loading('Proving ...', {id: 'provingToast'});
```
hackaugusto, line 118: The spinner is not spinning, it seems the proving is running on the rendering thread, so the whole webpage becomes unresponsive. I tested this with the fibonnaci example by changing from `1000` to `100000`.

I think the calls to the WASM code should be moved to a worker instead.

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -197,8 +207,13 @@ export default function CodingEnvironment(): JSX.Element {
   const verifyProgram = async () => {
     init().then(() => {
       disableDebug();
+      if (!proof) {
```
hackaugusto, line 210: instead of showing an error to the user, we should disable the `Verify` button, and only enable it one a proof is populated

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -221,15 +236,46 @@ export default function CodingEnvironment(): JSX.Element {
...
+      return;
+    }
+    toast.custom((t) => (
+      <div
+        className={`${
+          t.visible ? 'animate-enter' : 'animate-leave'
+        } max-w-full w-full bg-white shadow-lg rounded-lg flex ring-1 ring-black ring-opacity-5 overflow-hidden`}
+      >
+        <div className="flex-1 w-0 p-4 break-normal overflow-hidden">
+          <h1 className="text-lg font-medium text-gray-900"> This is your proof as Uint8Array </h1>
```
hackaugusto, line 255: `Uint8Array` is a JS specific thing that is unrelated to Miden, it is not useful for the user to know this.

```suggestion
          <h1 className="text-lg font-medium text-gray-900"> This is your proof </h1>
```

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -221,15 +236,46 @@ export default function CodingEnvironment(): JSX.Element {
...
+          {proof.toString()}
```
hackaugusto, line 256: Seeing the proof like this isn't useful, here are some suggestions:

- change from toast to a overlay with a `<code>` block with a scrollbar and show the complete proof using hex-encoding
- hex-encode the proof, and show only its beginning

In both cases, it would be useful to know the proof size in bytes

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -300,6 +349,13 @@ export default function CodingEnvironment(): JSX.Element {
           />
         </div>
       </div>
+      { proof ? 
+        <div className="flex">
+          <div className="absolute bottom-2 right-2 z-40">
+            <ActionButton label="Show Proof" onClick={showProof} />
```
hackaugusto, line 355: This button is hovering in a very odd position, it would make more sense to add it along side the `proof` and `verify` buttons instead, and make it clickable only after the proof is generated

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -134,14 +142,16 @@ export default function CodingEnvironment(): JSX.Element {
 "trace_len" : ${trace_len}
 }`);
         }
-        // Store the proof if >0 (empty proof is 0)
+        
+        toast.success(`Proving successful in ${Date.now() - start} ms`, {id: 'provingToast'});
```
hackaugusto, line 146: please also add the timing to the verify step

hackaugusto (changes requested):
looks good, requested some changes, doing the execution/proving on a worker should probably be done on another PR since that may not be a trivial change

hackaugusto:
btw, I'm not sure i understand the purpose of the shield, do we need it?

playground/src/pages/CodingEnvironment.tsx
Dominik1999: applied the formatter

playground/src/pages/CodingEnvironment.tsx
Dominik1999: Let's have a call about that, I don't understand directly

playground/src/pages/CodingEnvironment.tsx
Dominik1999: ok

playground/src/pages/CodingEnvironment.tsx
Dominik1999: good point

Dominik1999: done

playground/src/pages/CodingEnvironment.tsx
Dominik1999: done

playground/src/pages/CodingEnvironment.tsx
Dominik1999: done, check it out

playground/src/pages/CodingEnvironment.tsx
Dominik1999: done, now there are many buttons at the top

Dominik1999:
> 

Probably not. The less the better

hackaugusto (approved):
LGTM

Issue #89, opened 2023-03-15
Format Output in Rust not in Typescript


Dominik1999:
Once we have settled on what we want to show in the Debugging Output, we should format the output to a JSON in Rust. That way, in the front end we can show what we want.

At the moment, we return a structure `Outputs` from Rust land when the user executes Run or Prove.

https://github.com/0xPolygonMiden/examples/blob/8b836be8a24d2bf95e5c0fc805198f7c17d6f612/playground/miden-wasm/src/lib.rs#L17-L24

It would be good to be consistent with the Debugging feature and return a similar data structure for all features. We can then format the Outputs in JS land how we want.

PR #108, opened 2023-03-21
Imroving the debugging

### LINKED ISSUE #94: Improvements for the Debugging feature
This list is WIP. The goal is to improve the debugging feature of the playground. 

- [x] When the users clicks `Debug` the program should start and display clk=0
- [x] We need to be able to jump more than one clk (maybe one button for 10 and one for 100)?
- [x] `op` and `asmop` are quite confusing for users. It is hard to map them to Assembler Instructions (is that true?)
- [x] `fmp` is not super helpful to show
- [x] memory is not shown correctly

Dominik1999:
Closes #94 #89 

- The Debugging Output is being returned as a Struct
- The Debugging Output is being formatted in JS land
- A user can now jump 100 clock cycles back and forth with extra buttons


examples/fibonacci.masm
```diff
@@ -2,6 +2,8 @@
 
 begin
     repeat.1000
-        swap dup.1 add
+        swap
```
Dominik1999, line 5: Changed that for better readability. Almost no one understands the example looking at it for the first time

playground/miden-wasm/src/utils_debug.rs
```diff
@@ -140,57 +145,42 @@ impl DebugExecutor {
...
-    pub fn print_memory(&self) -> String {
-        let mem_result = String::new();
-        for (address, mem) in self.vm_state.memory.iter() {
-            format!("{}{}", mem_result, Self::print_memory_data(address, mem));
-        }
-        return mem_result;
+    /// converts a word to a tuple of 4 u64s.
+    /// this is necessary because wasm_bindgen does not support arrays of arrays.
+    fn word_to_ints(&self, word: &Word) -> [u64; 4] {
+        [word[0].as_int(), word[1].as_int(), word[2].as_int(), word[3].as_int()]
```
hackaugusto, line 173: nit: this is not using `self`, should probably be a standalone function.

playground/src/markdown/Explainer.md
```diff
@@ -76,28 +76,30 @@ You can create a program and run it. There will be no proof generated which is m
...
 
-* clk is the current (clock) cycle 
-* op is the current assembler operation 
-* asmop shows shows more info on the operation
-* fmp shows the free memory pointer (fmp), which is initially set to 2^30
+* clk is the current (clock) cycle
 * stack shows the current state ot the stack
-* memory shows (address, mem) pairs if used
+* miden operation is the current miden vm operation (to which assembler operations get translated)
+* assembly operation shows the currebt Miden assembler operation
```
hackaugusto, line 92: typo:
```suggestion
* assembly operation shows the current Miden assembler operation
```

playground/src/markdown/Explainer.md
```diff
@@ -76,28 +76,30 @@ You can create a program and run it. There will be no proof generated which is m
...
+* assembly num of cycles shows how long the currenct assembly operation takes
```
hackaugusto, line 93: typo:
```suggestion
* assembly num of cycles shows how long the current assembly operation takes
```

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -189,14 +191,14 @@ export default function CodingEnvironment(): JSX.Element {
...
+        setOutput(formatDebuggerOutput(debugOutput));
       } else {
-        const result = debugExecutor.execute(command);
-        setOutput(addNewlineAfterWhitespace(result));
+        const debugOutput : DebugOutput = debugExecutor.execute(command);
+        setOutput(formatDebuggerOutput(debugOutput));
       }
     } catch (error) {
-      setOutput("Error: Check the developer console for details.");
+      setOutput("Error: Check the developer AA console for details.");
```
hackaugusto, line 201:
```suggestion
      setOutput("Error: Check the developer console for details.");
```

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -260,9 +262,13 @@ export default function CodingEnvironment(): JSX.Element {
       {showDebug ? (
         <div className="flex justify-center pt-6">
```
hackaugusto, line 263: The buttons borders are not rounded to the right direction:

image

the left one should be `rounded-l-md`, the middle ones `rounded-none`, the right one `rounded-r-md`

playground/src/pages/CodingEnvironment.tsx
```diff
@@ -260,9 +262,13 @@ export default function CodingEnvironment(): JSX.Element {
...
           <DebugButton
-            icon="PPrevious"
+            icon="Start"
             onClick={() => executeDebug(DebugCommand.RewindAll)}
           />
+          <DebugButton
+            icon="PPrevious"
+            onClick={() => executeDebug(DebugCommand.Rewind, BigInt(100))}
```
hackaugusto, line 270: It is unclear what this does from the icons alone, please add a title to eac button, suggestions:

- `Go to start`
- `100 steps backwards`
- `step backwards`
- `step`
- `100 steps`
- `Go to end`

playground/src/utils/helper_functions.tsx
```diff
@@ -139,22 +141,41 @@ Did you prove the program first?`;
...
-  }
-  return result;
+  return Math.max(power, 1024);
+} 
+
+/**
+ * Helper function to format the Debugger output.
+ */
+export function formatDebuggerOutput(debugOutput: DebugOutput): string {
+  const output =
```
hackaugusto, line 159: nit: this looks like JSON, but it is not. Two possible improvements:

- Just print the JSON
- remove the quotes and show elements as a tidy table, i.e.:

```
clock: 5
stack: [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
instruction: push.1
operation: Some(Incr)
operation index: 2
operation count: 2
memory: 
```

playground/miden-wasm/src/utils_debug.rs
```diff
@@ -140,57 +145,42 @@ impl DebugExecutor {
...
-                    s
-                });
-        res.pop(); // removes unnecessary line-break
-        res
+        let memory: Vec<(u64, [u64; 4])> =
+            self.vm_state.memory.iter().map(|x| (x.0, self.word_to_ints(&x.1))).collect();
+
+        let output = DebugOutput {
+            clk: self.vm_state.clk,
+            op: format!{"{:?}", self.vm_state.op},
```
hackaugusto, line 159: 1. Let stick to the same naming as in the VM. The assembly code has instructions, the virtual machine executes operations.

So this should be `instruction`.

2. It seems `op` is a `Option`, so this formats like `None` or `Some(Inc)`. I think we change `DebugOutput.op` to be `Option<String>` and this line should be:

```rust
self.vm_state.op.map(|v| format!("{:?}" v))
```

Then the JS should handle the `None` case.

playground/miden-wasm/src/utils_debug.rs
```diff
@@ -140,57 +145,42 @@ impl DebugExecutor {
...
+            asmop_op: asmop.op().to_string(),
```
hackaugusto, line 160: since above we changed to instruction, this can be just `op` or `operation`.

playground/miden-wasm/src/utils_debug.rs
```diff
@@ -140,57 +145,42 @@ impl DebugExecutor {
...
+            asmop_num_cycles: asmop.num_cycles(),
```
hackaugusto, line 161: Every operation takes a single cycle, but an instruction can be composed of multiple operations. To make it clearer it is better to name this `number_of_ops` or `number_of_operations`.

playground/miden-wasm/src/utils_debug.rs
```diff
@@ -140,57 +145,42 @@ impl DebugExecutor {
...
+            asmop_cycle_idx: asmop.cycle_idx(),
```
hackaugusto, line 162: This can be a variant of `op_idx` or `operation_index`

playground/src/pages/CodingEnvironment.tsx
hackaugusto: btw, which version of the buttons is being used? this one, or the one in `playground/src/components/DebugButtons.tsx`?

hackaugusto (changes requested):
nice! left some comments

playground/miden-wasm/src/utils_debug.rs
```diff
@@ -14,17 +15,25 @@ pub struct DebugExecutor {
     vm_state: VmState,
 }
 
+// Debugging Output
```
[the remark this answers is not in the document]
Dominik1999: now we have a better comment

playground/miden-wasm/src/utils_debug.rs
Dominik1999: jup

playground/miden-wasm/src/utils_debug.rs
Dominik1999: Nice feedback, this is super helpful

So the VM has operations and Assembly Instructions (which is not quite consistently used when I look at `asmop`): which leads to the following table; 

```
clock: 3100
stack: [9,31,1,0,1,0,3,1,31,10,50,29,23,19,17,13,11,7,5,3,2,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
VM Operation: Some(MovUp3)
Assembly Instruction: gt
Instruction num of cycles: 18
Instruction cycle index: 13
Memory (Addr, Mem): [0]: [2, 0, 0, 0] 
                    [1]: [3, 0, 0, 0] 
```

playground/src/pages/CodingEnvironment.tsx
Dominik1999: Good point I moved the buttons down and squared them

playground/src/markdown/Explainer.md
```diff
@@ -76,28 +76,30 @@ You can create a program and run it. There will be no proof generated which is m
...
-* asmop shows shows more info on the operation
-* fmp shows the free memory pointer (fmp), which is initially set to 2^30
-* stack shows the current state ot the stack
-* memory shows (address, mem) pairs if used
+* Clock is the current (clock) cycle
+* Stack shows the current state ot the stack
+* VM Operation is the current miden vm operation (every Operation takes 1 cycle)
+* Assembly Instruction shows the current Miden assembler operation
+* Number of Operations shows cycle count for the instruction
+* Rel. Operation Index tells you at which relative cycle you are in the Instruction
```
hackaugusto, line 94: I believe that the assembly <-> operation relation is a bit confusing, and we should try to clarify it a bit better.

I would have the details about an operation together, since they are related. And the instruction first, since that is the thing the user will be changing, it seems to me like it is the most important item. (here and the output too)

In the docs I would also add a link to the Miden documentation. Something like this:

```suggestion
* Assembly Instruction: current Miden assembly instruction.
* Number of Operations: number of operations per instruction.
* Rel. Operation Index: current operation count.
* VM Operation: current Miden VM operation, every operation takes 1 cycle.

Additional documentation on how the VM executes its operations can be foound at the [Miden VM documentation page](https://0xpolygonmiden.github.io/miden-vm/design/programs.html)
```

Dominik1999: very good point, changed also the output accordingly

hackaugusto (approved):
LGTM

Issue #90, opened 2023-03-16
Simple UI cleanups for the playground


Dominik1999:
- [ ] Spacing consistency
- [ ] Size consistency
- [ ] Margin
- [ ] Only have our two colours (White and Polygon blue)
- [x] Font in the instruction table can also be black (from Andrey)
- [ ] Clicking links should open a new window 
- [ ] Header section takes too much real estate
- [x] Shorten the name of the playground
- [ ] The ActionButtons Run, Debug, Prove, Verify can be made "sticky"
- [ ] Add a button to always come back to top of the page
- [ ] Use Tailwind UI Component and simply copy our stuff in there (e.g. [here](https://tailwindui.com/components/marketing/elements/headers#component-9b3074c76a13e5151c82c2be12e52a6c))

Dominik1999:
done by @grjte in #115 and #116

PR #115, opened 2023-03-24
UI design changes

grjte:
This PR aims to partially address #90 by tidying up the UI a bit with:

- more consistent spacing & text sizes
- a new header and tab navigation that take up less space and open external links in new tabs
- brand colors
- responsiveness (except on the instruction table where this still needs to be fixed)
- backgrounds for the code windows
- sticky playground buttons on scroll

This PR does not include minor UX improvements, which I plan to do separately.

Note: this was a pretty quick pass, so I wasn't very diligent about TypeScript typing - we should revisit the codebase at some point and add proptypes & remove the "any" types.


hackaugusto (approved):
awesome, looks much better :)

The proof popup is below the header though, everything else looks great!

grjte:
> awesome, looks much better :)
> 
> The proof popup is below the header though, everything else looks great!

Thank you! I've left the proof popup for a follow-up PR, since I'm updating its design to a modal.

PR #116, opened 2023-03-24
Playground UX improvements

grjte:
This PR addresses the rest of #90 and improves the UX in the following ways

- new proof modal for displaying the whole proof
- disable buttons while the playground is executing or proving
- typescript types updated

Note: since execution is quite fast, maybe we shouldn't disable the buttons during execution and only do this during proving?


hackaugusto:
> Note: since execution is quite fast, maybe we shouldn't disable the buttons during execution and only do this during proving?

I believe the buttons are disabled because to verify the proof, first you have to prove to set the internal state. Is that what you meant?

hackaugusto (approved):
LGTM

grjte:
> I believe the buttons are disabled because to verify the proof, first you have to prove to set the internal state. Is that what you meant?

No, what I meant is that I made a change in this PR so that when you press "Run" or "Prove" all buttons get disabled while the VM is processing, then re-enabled afterwards. This improves the UX during proving, since it's very clear that something is happening, and you can't double-submit the request. During execution, this may not be necessary, since execution is fast. So this change actually causes the buttons to blink, and maybe it would be better to only apply this change to proving rather than to both.

PR #119, opened 2023-03-28
Update advice provider object names

frisitano:
As part of the recent changes to `miden-vm` the names of the objects contained within the advice provider have change and so have their associated counter parts in the `miden-cli`.  The `cli` now expects the objects in the input file to be named `operand_stack` and `advice_stack`.  This PR makes the change to the example input files. 

InputFile is now expected to be defined as [follows](https://github.com/0xPolygonMiden/miden-vm/blob/next/miden/src/cli/data.rs#L37-L41):
```rust
#[derive(Deserialize, Debug)]
pub struct InputFile {
    pub operand_stack: Vec<String>,
    pub advice_stack: Option<Vec<String>>,
}
```


grjte (approved):
Looks good, thank you!

Dominik1999 (approved):
Looks good to me, thanks. 

I am just confused as a Miden user. 

The docs say there are `public inputs` and `secret (non-deterministic) inputs`. 

https://0xpolygonmiden.github.io/miden-vm/intro/overview.html#inputs-and-outputs

Why don't we call those two input types in the input file `public` and `secret`?

It feels confusing to use `operand_stack` and `advice_stack` when I only read the docs. It is explained in the Design section of the docs. 

No need to change that now or to not approve this PR.

frisitano:
> Why don't we call those two input types in the input file `public` and `secret`?
> 
> It feels confusing to use `operand_stack` and `advice_stack` when I only read the docs. It is explained in the Design section of the docs.
> 
> No need to change that now or to not approve this PR.

I agree that this may be a bit confusing / complex but I don't think it can be avoided.  The terminology used here reflects the naming of the data structures used under the hood in the VM. To use them inside the VM requires awareness and understanding of the structures and how they work.

frisitano:
> I think it makes sense to update this when `0.5.0` of the vm is released, and to update the [dependency version](https://github.com/0xPolygonMiden/examples/blob/next/playground/miden-wasm/Cargo.toml#L11-L12) and the [explainer](https://github.com/0xPolygonMiden/examples/blob/5faf61372edc86b1451d7303a442f9bb80fd3d4d/playground/src/markdown/Explainer.md#L17-L22) and the JS code

This has been updated now in #122.  Could you please review so we can get this merged please.

PR #120, opened 2023-03-28
Advice provider example

frisitano:
This PR provides an example of how to define advice provider data and access it using miden assembly.


examples/advice_provider.inputs
```diff
@@ -0,0 +1,19 @@
+{
+    "operand_stack": [],
+    "advice_stack": ["1", "2", "3", "4"],
+    "advice_map": {
+        "0000000000000000000000000000000000000000000000000000000000000000": [5, 6, 7, 8],
+        "0000001000000000000000000000000000000000000000000000000000000000": [9, 10, 11, 12]
```
vlopes11, line 6: Do you think a test with hex format (i.e. `0xbeef`) will work here? If yes, we could add one entry with such format to demonstrate it is also possible.

examples/advice_provider.masm
```diff
@@ -0,0 +1,66 @@
...
+
+    # fetch the leaf
+    mtree_get
+
+    # assert the word is what we would expect ("0000000000000000000000000000000333000000000000000000000000000000")
+    push.0.216172782113783808.51.0 assert_eqw
+
+    # drop the root
+    dropw
+end
```
vlopes11, line 66: Optional nit: EOF line break

vlopes11 (approved):
Looks good! Two optional items

examples/advice_provider.inputs
frisitano: This wouldn't work.  The keys defined int he `.inputs` file must be 32 bytes long (64 hex characters).  We could maybe add some logic in the future to pad if this is not the case.

examples/advice_provider.masm
frisitano: updated

PR #122, opened 2023-03-29
Advice provider playground update

frisitano:
This PR achieves the following:
- Updates object names associated with `ProgramInputs` (`operand_stack`, `advice_stack`)
- Adds support for `advice_map` and `merkle_store`
- Refactors input and output validation such that errors are provided via the UI
- Refactors `miden-wasm` backend to return `Result<_, JsValue>` type and never panic
- Updated front end to parse errors returned from the `miden-wasm` backend and display them to the user

TODO:
- update typescript test suite to account for the logic above.  
- Add updates to epxlainer.

Call to action:  It would be great if we could collaborative test the experience to ensure that the tests which have implemented are stabled.


Dominik1999 (changes requested):

playground/miden-wasm/src/lib.rs
```diff
@@ -328,9 +334,9 @@ fn test_verify_program() {
         "trace_len": 1024
     }"#;
 
-    let proof = expected_test_proof::EXPECTED_PROOF_BYTES.to_vec();
+    let proof = prove_program(asm, input_str).unwrap();
 
-    let result = verify_program(asm, input_str, output_str, proof);
+    let result = verify_program(asm, input_str, output_str, proof.proof.unwrap()).unwrap();
```
grjte, line 339: nit: `proof.proof.unwrap()` is a bit strange. Maybe `proof` above should be renamed to something else (`output`, `result`, etc)?

grjte, line 339: nit: `proof.proof.unwrap()` is a bit strange. Maybe `proof` above should be renamed to something else (`output`, `result`, etc)?

Dominik1999:
@frisitano thanks a lot. That looks great. I left some nits and questions. And in general, I am not sure if we should allow those complex inputs like `merkle_store` in the playground. There is certainly no harm in doing so, but because you cannot create those data structures in the playground, I would expect users to use the CLI anyway for this.

And we need to change all the other examples as well because they don't work now anymore. I can do that after your PR is merged

playground/src/utils/helper_functions.tsx
```diff
@@ -21,121 +22,175 @@ export interface checkedData {
...
   }
 
+  // Parse the json output
   let jsonOutput!: JSON;
-
   try {
     jsonOutput = JSON.parse(jsonString);
-  } catch (e: any) {// eslint-disable-line @typescript-eslint/no-explicit-any
+  } catch (e: any) {
+    // eslint-disable-line @typescript-eslint/no-explicit-any
```
grjte, line 181: does this still work to disable linting for the line with the `any` type? I thought it had to be on the same line (or on the line before it using `eslint-disable-next-line` instead of `eslint-disable-line`)

It would be better to just use an `Error` type instead of `any`, but that doesn't have to happen in this Pr

playground/miden-wasm/src/utils_input.rs
```diff
@@ -14,11 +16,195 @@ pub struct Outputs {
...
+/// - advice_stack
+/// - advice_map
+/// - merkle_store
+#[derive(serde::Deserialize)]
 pub struct InputFile {
-    pub stack_init: Option<Vec<String>>,
-    pub advice_tape: Option<Vec<String>>,
+    /// String representation of the initial operand stack, composed of chained field elements.
+    pub operand_stack: Vec<String>,
+    /// Opitonal string representation of the initial advice stack, composed of chained field
```
grjte, line 51: nit: typo in optional

playground/miden-wasm/src/lib.rs
```diff
@@ -303,10 +313,6 @@ fn test_prove_program() {
 
     // we expect a proof of []
     assert!(output.proof.is_some());
-    assert_eq!(
```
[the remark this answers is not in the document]
frisitano: I think it makes sense to test that we get a proof which is what we do above with `output.proof.is_some()`. I have then modified the verify function such that the proof that is returned is verified so we have internal consistency.  The persisted proof that we were using before became stale and doesn't work with the latest version of miden-vm.

playground/miden-wasm/src/utils_input.rs
```diff
@@ -14,11 +16,195 @@ pub struct Outputs {
...
+    /// String representation of a merkle tree.  The merkle tree is represented as a vector of
+    /// 32 byte hex strings where each string represents a leaf in the tree.
+    #[serde(rename = "merkle_tree")]
+    MerkleTree(Vec<String>),
+    /// String representation of a sparse merkle tree. The sparse merkle tree is represented as a
+    /// vector of tuples where each tuple consists of a u64 node index and a 32 byte hex string
+    /// representing the value of the node.
+    #[serde(rename = "sparse_merkle_tree")]
+    SparseMerkleTree(Vec<(u64, String)>),
+}
```
grjte, line 35: nit: I would use "Merkle" instead of "merkle" everywhere

playground/src/components/CodingEnvironment/MidenInputs.tsx
```diff
@@ -11,7 +11,7 @@ type MidenInputsProps = {
 const MidenInputs = (props: MidenInputsProps): JSX.Element =>
     <div className="min-w-0 flex-1 box-border">
         <EditorLabel label="inputs" />
-        <Editor height="150px" {...props} />
+        <Editor height="400px" {...props} />
```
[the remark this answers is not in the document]
frisitano: I think 150px limits ones ability to create / edit an advice provider that has significant data input.  Happy to change it back if you think that's best.  Let wait for some feedback from others.

examples/advice_provider.masm
```diff
@@ -2,6 +2,9 @@
 # miden assembly. Please see the associated `advice_provider.inputs` file for the definition of
```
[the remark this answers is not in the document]
frisitano: The inputs don't really have any logical meaning.  It's just dummy data so we can provide an example of how to access it.  I can change this if needed.  What would you expect to see?

playground/src/utils/helper_functions.tsx
```diff
@@ -21,121 +22,175 @@ export interface checkedData {
...
+        if (typeof value === 'string') {
+          return Number.isInteger(parseInt(value ?? ''));
+        }
+        return false;
+      }
+    );
+};
+
+// Schema for merkle_store
+const merkleStoreParser = yup
```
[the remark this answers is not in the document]
frisitano: The user would have to create the structures manually for both the cli and for the playground. This isn't a great developer experience and I think we should look to build some support for this in the future.  However, merkle data structures are a foundational data structure that form the basis of a lot of applications so I think it's quite important that we support this.  I don't think we should constrain the scope of what's possible in the plagyround.

playground/src/utils/helper_functions.tsx
```diff
@@ -21,121 +22,175 @@ export interface checkedData {
...
+  )
+  .notRequired();
+
+// Schema for the whole input
+const inputSchema = yup
+  .object()
+  .shape({
+    operand_stack: yup
+      .array()
+      .of(integerStringArraySchema('operand_stack'))
```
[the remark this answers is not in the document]
frisitano: thank you!  I will look into this.

playground/src/components/CodingEnvironment/MidenInputs.tsx
grjte, line 14: I think 400px might be too much, since it pushes the large code window quite far down the page on my laptop screen. For most of the simple cases, there isn't that much going on in the inputs, and 400px is more than needed for the outputs. Ideally we would have the option to expand these or keep them smaller, but that's a UI update for the future. I probably wouldn't enlarge it past 250px

playground/src/components/CodingEnvironment/MidenOutputs.tsx
```diff
@@ -75,7 +75,7 @@ const MidenOutputs = ({ showDebug, debugExecutor, ...props }: MidenOutputsProps)
                     </div>
                 ) : null}
             </div>
-            <Editor height="150px" {...props} />
+            <Editor height="400px" {...props} />
```
grjte, line 78: Same comment here about the size - maybe 250px instead?

grjte (commented):
Thank you! I left a couple of minor non-blocking comments. I'm not able to run this up locally unfortunately (due to network issues, I think), so I can't really judge fully & therefore won't approve at this point.

Regarding my points about the size of inputs/outputs - it's hard for me to really judge this without having been able to try out the new advice inputs, but I think the page looks better when they're smaller, and most of the examples don't use the complex advice inputs

playground/src/utils/helper_functions.tsx
```diff
@@ -21,121 +22,175 @@ export interface checkedData {
...
-/**
- * Helper function to iterate over the inputs and check if they are valid.
- */
-export function checkFields(jsonInput: JSON): checkedData {
-  let returnValue: checkedData = { isValid: true, errorMessage: "" };
-
-  for (const key in jsonInput) {
-    if (key === "trace_len") {
-      continue;
+// Parser for advice_map
```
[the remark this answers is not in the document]
frisitano: yeah it's pretty useful.  It's quite commonly used in association with [`formik`](https://formik.org/docs/guides/typescript) where `yup` provides the input validation.

playground/miden-wasm/src/lib.rs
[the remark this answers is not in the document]
grjte: I prefer the new approach

playground/src/utils/helper_functions.tsx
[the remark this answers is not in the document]
frisitano: Thank you for the review!

> And in general, I am not sure if we should allow those complex inputs like `merkle_store` in the playground. There is certainly no harm in doing so, but because you cannot create those data structures in the playground, I would expect users to use the CLI anyway for this.

I think it's quite important that we provide support for merkle data structures as they form the basis of a lot of useful applications as mentioned in my response above.  However, I do agree that we need to build better support for actually building the merke store inputs. 

> And we need to change all the other examples as well because they don't work now anymore. I can do that after your PR is merged

I have a PR lined up for this:
https://github.com/0xPolygonMiden/examples/pull/119/files

playground/src/utils/helper_functions.tsx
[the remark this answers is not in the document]
grjte: I agree with @frisitano - the CLI and the playground currently have the same problem, so I don't see a reason to limit the functionality in the playground. If people decide they'd rather use the CLI, they can still do so. The problem of finding a better way for creating these structures is something we should tackle in the future for both CLI & playground.

Btw - I think we should have an obvious mention of the CLI/Debugger/REPL somewhere in the playground. Wdyt? Maybe the simplest thing would be to add another link in the navigation that says "CLI/REPL" and links to the user docs page for those? That page is a bit suboptimal though since it doesn't have the github link. Maybe we should update the docs to have instructions to clone the miden-vm repo (or at least provide the link to it). In that case, we would probably want to link to the docs that are deployed from next rather than the ones on wiki.polygon.technology

examples/advice_provider.inputs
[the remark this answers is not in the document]
Dominik1999: Nope, let's do that. Was just not sure if you were aware

examples/advice_provider.masm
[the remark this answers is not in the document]
Dominik1999: When I was reading `Please see the associated `advice_provider.inputs` file for the definition of the advice_provider inputs` I was expecting to get more info in the file about what these data fields are. But that was not in the file. 

We don't really explain what an `advice_map` or a `merkle_store` is. Can we link to the docs in that comment?

Dominik1999:
> Thank you for the review!
> 
> > And in general, I am not sure if we should allow those complex inputs like `merkle_store` in the playground. There is certainly no harm in doing so, but because you cannot create those data structures in the playground, I would expect users to use the CLI anyway for this.
> 
> I think it's quite important that we provide support for merkle data structures as they form the basis of a lot of useful applications as mentioned in my response above. However, I do agree that we need to build better support for actually building the merke store inputs.
> 
> > And we need to change all the other examples as well because they don't work now anymore. I can do that after your PR is merged
> 
> I have a PR lined up for this: https://github.com/0xPolygonMiden/examples/pull/119/files

I am convinced, let's do that.

playground/miden-wasm/src/utils_input.rs
frisitano: I'm not sure about this.  I think it would look inconsistent when defining the `.inputs` file.  Here's an example of how it would look.  Happy to change it if you prefer it.

```json
{
    "operand_stack": ["1"],
    "advice_stack": ["1", "2", "3", "4"],
    "advice_map": {
        "0000000000000000000000000000000000000000000000000000000000000000": [5, 6, 7, 8],
        "0000001000000000000000000000000000000000000000000000000000000000": [9, 10, 11, 12]
    },
    "merkle_store": [
        {"MerkleTree": [
            "0000000000000000000000000000000000000000000000000000000000000000", 
            "0000000000000000000000000000000000000000000000000000000000000000", 
            "0000000000000000000000000000000000000000000000000000000000000000", 
            "0000000000000000000000000000000000000000000000000000000000000000"]},
        {"SparseMerkleTree": [
            [1, "0000000000000000000000000000000333000000000000000000000000000000"],
            [3, "0000000000000000000000000000000222000000000000000000000000000000"]
        ]}
    ]
}
```

frisitano: Not sure if you were referring to the comment or the serde renaming.

playground/src/components/CodingEnvironment/MidenInputs.tsx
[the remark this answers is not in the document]
frisitano: changed to 250px.

playground/miden-wasm/src/lib.rs
frisitano: changed

playground/miden-wasm/src/utils_input.rs
frisitano: fixed

playground/src/components/CodingEnvironment/MidenInputs.tsx
frisitano: changed to 250px

playground/miden-wasm/src/utils_input.rs
grjte: Oh yeah, I just meant in the comments. It was a very nitty nit :)

PR #125, opened 2023-03-31
Update advice inputs

frisitano:
This PR updates the advice inputs for the miden examples such that they are compatible with the new version of miden. This is a copy of #119 which is blocked but we need to get this delivered ASAP.


Dominik1999 (approved):
lgtm thanks

PR #130, opened 2023-03-31
Change `overflow_addrs` to being optional

frisitano:
`overflow_addrs` should be `notRequired()` as we do not always need it to verify a proof.  This PR modifies the `yup` schema to reflect this.  This impacts proof verification and can be recreated using the `advice_provider` example.

### DIFF
--- a/playground/src/utils/helper_functions.tsx
+++ b/playground/src/utils/helper_functions.tsx
@@ -158,7 +158,7 @@ export const checkInputs = (
 
 const outputSchema = yup.object().shape({
   stack_output: yup.array().of(yup.number().integer().min(0)).required(),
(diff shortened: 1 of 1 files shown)


Dominik1999:
I think the outputs are only going to be checked when the user wants to verify, right?

So do we need an `overflow_addrs` for every verification?

frisitano:
> I think the outputs are only going to be checked when the user wants to verify, right?
> 
> So do we need an `overflow_addrs` for every verification?

Yes that's correct and in some instances, like the advice_provider example, it's not needed for verification.  This PR changes the requirement of `overflow_addrs` such that it is optional.

bobbinth (approved):

PR #131, opened 2023-03-31
feat: adding the breakpoint for debugging

Dominik1999:
I am adding the `breakpoint` to the Playground. When the user adds `breakpoint` in the code and runs the debugger, it stops at the breakpoint.


frisitano (changes requested):
I just remembered that we should probably add some documentation about this to the `Explainer`.

frisitano (approved):
Looks great, thanks!

PR #132, opened 2023-03-31
Add examples & improve dropdown UI

grjte:
This PR does the following:

- adds a new example for using the standard libary
- updates the playground to use examples from the next branch (we can revert this when we release this version)
- updates the existing examples to have a consisting naming scheme with "_" instead of "-"
- updates the dropdown to show spaces instead of "_"
- fixes an error with verification of proofs that didn't output any `overflow_addrs`
- adds the rest of the examples to the playground, since they seem to be working now


playground/src/pages/CodingEnvironment.tsx
```diff
@@ -142,19 +142,12 @@ export default function CodingEnvironment(): JSX.Element {
         const start = Date.now();
         const { stack_output, trace_len, overflow_addrs, proof }: Outputs =
           prove_program(code, inputs);
-        const overflow = overflow_addrs ? overflow_addrs.toString() : null;
-        if (overflow) {
-          setOutput(`{
+        const overflow = overflow_addrs ? overflow_addrs.toString() : "[]";
```
hackaugusto, line 145: nit:

```suggestion
        const overflow = overflow_addrs.toString();
```

or 

```suggestion
        const overflow = overflow_addrs ? overflow_addrs.toString() : "";
```

as it is, we would have the output set to `[[]]`, which is probably not the intention. (I haven't check what type is `overflow_addrs` to approach is best)

hackaugusto (approved):
thanks! I left just one nit about the overflow output, all else seems great!

PR #133, opened 2023-03-31
Add examples & improve dropdown UI

grjte:
This PR supersedes #132 and does the following:

-  adds a new example for using the standard libary
- updates the existing examples to have a consisting naming scheme with "_" instead of "-"
- updates the dropdown to show spaces instead of "_"
- adds the rest of the examples to the playground, since they seem to be working now


hackaugusto (approved):
Looks great, thanks!

PR #134, opened 2023-04-09
new_example: greatest common divisor

py-zoid:
Adding the trivial example for calculating the greatest common divisor aka highest common factor. The example contains usage of inputs from the operand stack, 32-bit conditional operations, 32-bit arithmetic operations and conditional stack manipulation.


Dominik1999 (approved):
This is pretty amazing! Thanks for submitting. It works quite well

Also nice comments

PR #135, opened 2023-04-17
Fibonacci example fix

Overcastan:
Current version of Fibonacci example computes not 1000th, but 1002nd Fibonacci number. 
This PR changes this example to computation of 1001st Fibonacci number by putting only `[1]` to the input. This approach allows to remain `1000` as cycles number, changing only stack and a comment.


grjte (changes requested):
Looks good, thank you! I think are some docs that need updating as well. At least the README of the examples repo needs to be changed, but can you check everywhere to be sure?

Overcastan:
Looks like main README was the only place left to fix. `Fibonacci` also mentioned in the Benchmarking CLI but without specifying what number we are counting, so there is no need to edit anything there.

grjte (approved):
Looks good, thank you!

PR #145, opened 2023-07-13
Miden transaction prover 

frisitano:
This PR introduces WASM bindgen endpoints to support proving miden transactions in wasm.  This is a work in progress.


Dominik1999 (approved):
Approving now, I need it by tomorrow

PR #148, opened 2023-07-18
Refactor verify_transaction api

frisitano:


playground/miden-wasm/src/transaction.rs
Dominik1999, line 1: I like that it is a separate file

Dominik1999 (approved):
lgtm, no thorough review

PR #149, opened 2023-10-04
Upgrading to Miden v0.6 and adding Merkle Inclusion benchmark

Dominik1999:
I added a new example and upgraded the `benchmarking-cli` to Miden 0.6. 

The new example is basically a Merkle Inclusion Proof for 1000 merkle paths.


benchmarking-cli/src/utils_program.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct MidenProgram {
+    pub assembler: Assembler,
+    pub masm_code: String,
+    pub debug: bool,
+    pub program: Option<Program>,
+    pub program_info: Option<ProgramInfo>,
+}
+
+pub const DEBUG_OFF: bool = false;
+//pub const DEBUG_ON: bool = true;
```
hackaugusto, line 13: Suggestion: This allows the caller site to be more readable, `MidenProgram::new(prog, DEBUG_OFF);` instead of `MidenProgram::new(prog, false);`. But that is only the case if the user knows about the constant. Instead if you use:

```rust
enum Debug {
  On,
  Off,
}
```

We always get the nicer readability.

benchmarking-cli/src/utils_program.rs
```diff
@@ -0,0 +1,48 @@
...
+impl MidenProgram {
+    pub fn new(code_as_str: &str, debug: bool) -> Self {
+        Self {
+            assembler: Assembler::default(),
+            masm_code: code_as_str.to_string(),
+            debug,
+            program: None,
+            program_info: None,
+        }
+    }
```
hackaugusto, line 24: It's nice to support both `&str` and `String`. This could be done with:

```suggestion
    pub fn new<S: AsRef<str>>(code_as_str: S, debug: bool) -> Self {
        Self {
            assembler: Assembler::default(),
            masm_code: code_as_str.as_ref().to_string(),
            debug,
            program: None,
            program_info: None,
        }
    }
```

benchmarking-cli/src/utils_input.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+/// Miden Inputs plus Outputs that are used as inputs for the verifier.
+pub struct Inputs {
+    pub stack_inputs: StackInputs,
+    pub advice_provider: MemAdviceProvider,
+    pub stack_outputs: StackOutputs,
+}
+
+/// We need to implement the default trait for the Inputs struct.
```
hackaugusto, line 224: Is this a `TODO`?

benchmarking-cli/src/utils_input.rs
```diff
@@ -0,0 +1,264 @@
...
+impl Inputs {
+    pub fn new() -> Self {
+        Self {
+            stack_inputs: StackInputs::new(vec![Felt::ZERO]),
```
hackaugusto, line 228: Why is there a single `Zero` input? Shouldn't this be just `vec![]`?

benchmarking-cli/src/utils_input.rs
```diff
@@ -0,0 +1,264 @@
...
+        hex::decode_to_slice(word_hex, &mut word_data)
+            .map_err(|e| format!("failed to decode `Word` from hex {word_hex} - {e}"))?;
+        let mut word = Word::default();
+        for (i, value) in word_data.chunks(8).enumerate() {
+            word[i] = Felt::try_from(value).map_err(|e| {
+                format!("failed to convert `Word` data {word_hex} (element {i}) to Felt - {e}")
+            })?;
+        }
+        Ok(word)
+    }
```
hackaugusto, line 199: A few changes:

- This should handle the case were `word_hex` has an invalid number of bytes (either too many or too little).
- These methods that are not using `self`, are not constructors, and are not coming from a trait, should IMO be standalone functions.

---

Nit: It would also be nice to have this on `crypto` repo, along side this:

https://github.com/0xPolygonMiden/crypto/blob/18302d68e012fb008b5318da47e1525e219b97a8/src/utils/mod.rs#L24-L36

(probably better to open a separate PR and not block this one on that)

benchmarking-cli/src/utils_input.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub advice_stack: Option<Vec<String>>,
+    /// Optional map of 32 byte hex strings to vectors of u64s representing the initial advice map.
+    pub advice_map: Option<BTreeMap<String, Vec<u64>>>,
+    /// Optional vector of merkle data which will be loaded into the initial merkle store. Merkle
+    /// data is represented as 32 byte hex strings and node indexes are represented as u64s.
+    pub merkle_store: Option<Vec<MerkleData>>,
+}
+
+/// Helper methods to interact with the input file
+impl InputFile {
```
hackaugusto, line 62: Question: We do the same on the `miden` binary, no? Can't we use that code instead of re-implementing it?

hackaugusto (approved):
thanks! overall looks good, I left some comments to be addressed, once they are done I think we can merge this.

examples/merkle_inclusion.masm
```diff
@@ -0,0 +1,17 @@
...
+        dup.1 padw drop swap.3 
+        # => [i, 0, 0, 0, d, i, R,...]
+
+        # verify merkle path for node i
+        mtree_verify
+        # => [i, 0, 0, 0, d, i, R ...]
+
+        push.1 add swap.5 dropw
+        # => [d, i+1, R, ...]
+    end
```
bobbinth, line 16: Is there a reason to do both `padw` and then `dropw` inside the loop? I think the following could work just as well:
```
begin
    repeat.1000
        # stack = [i, 0, 0, 0, d, i, R, ...]

        # verify merkle path for node i
        mtree_verify
        # => [i, 0, 0, 0, d, i, R ...]

        add.1 swap.5 add.1
        # => [i+1, 0, 0, 0, d, i+1, R, ...]
    end
end
```

bobbinth (commented):
Looks good! Thank you! Just a couple of small comments from my end.

benchmarking-cli/src/utils_input.rs
bobbinth: I think this is probably copied directly from there. I agree that would be good to use a common component - but this is probably for a different PR. Let's create an issue for this.

benchmarking-cli/src/utils_program.rs
Dominik1999: this is nice and abstract

benchmarking-cli/src/utils_program.rs
Dominik1999: I removed it now, I think, no one wants to benchmark in DebugMode

benchmarking-cli/src/utils_input.rs
Dominik1999: > This should handle the case were word_hex has an invalid number of bytes (either too many or too little).

Will be done in the common file, actually we should use this here 

https://github.com/0xPolygonMiden/miden-vm/blob/5c5749549d5cf1ccb4a6b9c5cf38427a41537145/miden/src/cli/data.rs#L268-L280

hackaugusto (approved):
LGTM

PR #150, opened 2023-10-05
Add Discrete Fourier Transform example

plafer:
Implements the [Discrete Fourier Transform](https://en.wikipedia.org/wiki/Discrete_Fourier_transform_over_a_ring) over the input list. Basically, this is the non-efficient version of the Number Theoretic Transform (NTT).

You can verify the correctness of the program by comparing with the Python implementation:

```py
# Adapted from https://www.geeksforgeeks.org/python-number-theoretic-transformation/
from sympy import ntt 
  
seq = [15, 21, 13, 44] 
  
prime_no = 2**64 - 2**32 + 1
  
# ntt 
transform = ntt(seq, prime_no) 
print ("NTT : ", transform)

# NTT :  [93, 18440270144950239235, 18446744069414584284, 6473924464345090]
```

Conceptually though, this Miden implementation is closer to

```py
from math import log2

PRIME = 2**64 - 2**32 + 1
TWO_ADIC_ROOT_OF_UNITY = 7277203076849721926

def get_root_of_unity(n):
    power = (1 << (32 - n)) % PRIME
    return pow(TWO_ADIC_ROOT_OF_UNITY, power, PRIME)

def f_k(seq, k, alpha):
    alpha_k = pow(alpha, k, PRIME)
    return sum([(seq[i] * pow(alpha_k, i, PRIME)) for i in range(len(seq))]) % PRIME

def ntt(seq, alpha):
    return [f_k(seq, k, alpha) for k in range(len(seq))]

seq = [15, 21, 13, 44]
alpha = get_root_of_unity(int(log2(len(seq))))

print(ntt(seq, alpha))
```

### DIFF
--- a/examples/dft.masm
+++ b/examples/dft.masm
@@ -27,6 +27,56 @@ const.INPUT_ARR_LOC=0
 # Location at 3 * 2^30, the first address "with no special meaning" in the root context
 const.RESULT_ARR_LOC=3221225472
 
+#! Computes log base 2. Only supports {1, 2, 4, 8, 16}.
+#! Input: [n]
+#! Output: [log2(n)] if n is supported, otherwise halts
+proc.log2
+    dup                                        # stack: [n, n]
+    eq.1                                       # stack: [0 or 1, n]
+    if.true
+        # n == 1
+        drop                                   # stack: []
+        push.0                                 # stack: [0]
+    else
+        dup                                    # stack: [n, n]
+        eq.2                                   # stack: [0 or 1, n]
+        if.true
+            # n == 2
+            drop                               # stack: []
+            push.1                             # stack: [1]
+        else
+            dup                                # stack: [n, n]
+            eq.4                               # stack: [0 or 1, n]
+            if.true
+                # n == 4
+                drop                           # stack: []
+                push.2                         # stack: [2]
+            else
+                dup                            # stack: [n, n]
+                eq.8                           # stack: [0 or 1, n]
+                if.true
+                    # n == 8
+                    drop                       # stack: []
+                    push.3                     # stack: [3]
+                else
+                    dup                        # stack: [n, n]
+                    eq.16                      # stack: [0 or 1, n]
+                    if.true
+                        # n == 16
+                        drop                       # stack: []
+                        push.4                     # stack: [4]
+                    else
+                        # n unsupported; halt
+                        push.1
+                        assertz
+                    end
+                end
+            end
+        end
+
+    end
+end
+
 #! Computes the primitive root of unity for a subgroup of size 2^n.
 #!
 #! Input: [n], the size of the input list (must be <= 32)
@@ -65,29 +115,27 @@ proc.f_k.5
     push.1                                     # push 1 to enter the loop
     while.true
         # compute `root_of_unity^(jk)`
-        loc_load.1                             # load root_of_unity
-        loc_load.0                             # load k
-        loc_load.3                             # load j
-        mul                                    # compute j*k
-        exp                                    # compute root_of_unity^(jk)
-        mul                                    # compute v{j} * root_of_unity^(jk)
+        loc_load.1                             # stack: [root_of_unity, v{j}, ..., v{n-1}]
+        loc_load.0                             # stack: [k , root_of_unity, v{j}, ..., v{n-1}]
+        loc_load.3                             # stack: [j, k , root_of_unity, v{j}, ..., v{n-1}]
+        mul                                    # stack: [j*k , root_of_unity, v{j}, ..., v{n-1}]
+        exp                                    # stack: [root_of_unity^(jk), v{j}, ..., v{n-1}]
+        mul                                    # stack: [v{j} * root_of_unity^(jk), v{j+1}, ..., v{n-1}]
 
         # update result
-        loc_load.4                             # load result
-        add                                    # sum previous partial result with v{j} * root_of_unity^(jk)
-        loc_store.4                            # store partial result
+        loc_load.4                             # stack: [result, v{j} * root_of_unity^(jk), v{j+1}, ..., v{n-1}]
+        add                                    # stack: [result + v{j} * root_of_unity^(jk), v{j+1}, ..., v{n-1}]
+        loc_store.4                            # stack: [v{j+1}, ..., v{n-1}]
 
-        # Stack here: [v{j+1}, ..., v{n-1}, ...]
-
-        # Update j
-        loc_load.3                             # load j
-        add.1                                  # increment j by 1
-        dup                                    # dup before store to keep j+1 on stack
-        loc_store.3                            # store j+1
+        # Increment j
+        loc_load.3                             # stack: [j, v{j+1}, ..., v{n-1}]
+        add.1                                  # stack: [j+1, v{j+1}, ..., v{n-1}]
+        dup                                    # stack: [j+1, j+1, v{j+1}, ..., v{n-1}]
+        loc_store.3                            # stack: [j+1, v{j+1}, ..., v{n-1}]
 
         # Check if we're done looping
-        loc_load.2                             # load n. stack: [n, j+1, ...]
-        neq                                     # Check if j+1 != n; if so, we continue. Else, we're done.
+        loc_load.2                             # stack: [n, j+1, v{j+1}, ..., v{n-1}]
+        neq                                    # stack: [1 or 0, v{j+1}, ..., v{n-1}]
     end
 
     # Return result
@@ -233,5 +281,8 @@ proc.main.3
 end
 
 begin
+    # TODO:
+    # fix f_k for k>0
+    # fix "unreachable" error
     exec.main
 end
--- a/examples/dft.masm
+++ b/examples/dft.masm
@@ -132,10 +132,15 @@ proc.store_array
 end
 
 #! Retrieves an array previously stored with `proc.store_array`.
-#! Input: None
+#! Input: [array_loc], the memory location to retrieve the array
 #! Output: [n, v0, ..., v{n-1}]
-proc.retrieve_array
-    mem_load.ARR_LOC                          # stack: [n, ...]
+#!
+#! Locals
+#! 0: array_loc
+proc.retrieve_array.1
+    dup                                       # stack: [array_loc, array_loc, ...]
+    loc_store.0                               # stack: [array_loc, ...]
+    mem_load                                  # stack: [n, ...]
 
     # Prepare stack for loop
     push.0.1                                  # pushes i=0, followed by 1 so that we enter the loop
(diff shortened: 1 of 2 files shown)


examples/dft.masm
```diff
@@ -0,0 +1,288 @@
...
+# [0, 2^30) is global memory, so we store it at the beginning.
+const.INPUT_ARR_LOC=0
+
+# Location at 3 * 2^30, the first address "with no special meaning" in the root context
+const.RESULT_ARR_LOC=3221225472
+
+#! Computes log base 2. Only supports {1, 2, 4, 8, 16}.
+#! Input: [n]
+#! Output: [log2(n)] if n is supported, otherwise halts
+proc.log2
```
Dominik1999, line 33: This is cool, but basically you hardcode the result of the log2(n) operation, right? 

We could also compute the base-2 logarithm (log2) of an integer "n" and use a simple algorithm that repeatedly divides "n" by 2 until "n" becomes 1, while keeping track of the number of divisions. The number of divisions will be equal to the log2 of "n". Basically we can only compute the log2 of even numbers `<2^{32}`. 

```
proc.log2    
    # it should fail if the provided number >= 2^{32}
    u32test assert
    
    # if should faul if the provided number is odd
    dup push.2 u32checked_mod assertz

    # now we know that the number is even and < 2^{32}
    # so we can have a while loop and divide by 2 until we reach 1
    
    # create a counter
    push.0
    # get the loop going
    push.1

    while.true
        swap u32checked_div.2
        # get the counter and add + 1
        swap add.1 swap
        # check if the result != 1, then we continue in the loop
        dup movup.2 swap neq.1
    end

end
```

I am sure this code can be optimized.

examples/dft.masm
```diff
@@ -0,0 +1,288 @@
...
+end
+
+#! Computes the primitive root of unity for a subgroup of size 2^n.
+#!
+#! Input: [log2_n], log2 of the size of the input list (must be <= 32)
+#! Output: [root_of_unity], where `root_of_unity` is the primitive root of unity of order 2^n
+#!
+#! Equivalent to Winterfell's implementation: 
+#! https://github.com/facebook/winterfell/blob/4543689f73a2efb9d30927535de7b4efe7e1802d/math/src/field/traits.rs#L254-L259
+proc.get_root_of_unity
```
Dominik1999, line 87: Nice use of constants

examples/dft.masm
```diff
@@ -0,0 +1,288 @@
...
+
+    # Stack here: [v0, ..., v{n-1}, ...]
+
+    push.1                                     # push 1 to enter the loop
+    while.true
+        # compute `root_of_unity^(jk)`
+        loc_load.1                             # stack: [root_of_unity, v{j}, ..., v{n-1}]
+        loc_load.0                             # stack: [k , root_of_unity, v{j}, ..., v{n-1}]
+        loc_load.3                             # stack: [j, k , root_of_unity, v{j}, ..., v{n-1}]
+        mul                                    # stack: [j*k , root_of_unity, v{j}, ..., v{n-1}]
```
Dominik1999, line 123: only `j` changes with every iteration, right? Can you load the other values out of the loop and the multiply with what you need?

Dominik1999 (commented):
This is amazing, thanks a lot.

I did a first round, but I need more time to wrap my head around it

examples/dft.masm
Dominik1999: thinking about it, it's not 100% accurate. We don't need to check whether the number is odd or even, we would need to check if it is a power of 2 to ensure we get integers back

Dominik1999: Or we should only check if the provided number is a power of 2. And we can do that with bitwise comparison

examples/dft.masm
```diff
@@ -0,0 +1,288 @@
...
+    loc_store.3                                # Set j = 0
+    push.0                                     # Prepare result = 0
+    loc_store.4                                # Set result = 0
+
+    # Stack here: [v0, ..., v{n-1}, ...]
+
+    push.1                                     # push 1 to enter the loop
+    while.true
+        # compute `root_of_unity^(jk)`
+        loc_load.1                             # stack: [root_of_unity, v{j}, ..., v{n-1}]
```
Dominik1999, line 120: I didn't get it the first time :), can you add a `j = 0` on this comment, then its clearer why `[v0, ...]` becomes `[v{j}, ...]`

examples/dft.masm
```diff
@@ -0,0 +1,288 @@
...
+#! Input: [k, root_of_unity, n, v_0, ..., v_{n-1}, ...]
+#! Output: [f_k, ...]
+#!
+#! Locals
+#! 0: k
+#! 1: root_of_unity
+#! 2: n (length of the list)
+#! 3: j (counter)
+#! 4: result (partial result of the computation)
+proc.f_k.5
```
Dominik1999, line 106: This is elegant. 

You could think of only storing `k, n, root_of_unity`. Only `j` and the `result` change with every iteration. But actually not sure if this would help. 

Nice implementation and nice comments!

Dominik1999 (commented):
Super nice example. I left a few comments but overall, its really nice. And I learned some things about the Discrete Fourier Transform (DFT). 

One general comment, we now try to follow a common commenting style in MASM, see https://github.com/0xPolygonMiden/miden-base/blob/main/miden-lib/asm/faucets/basic.masm#L39C1-L39C1 where we always 

- explain a function or line of instructions above it, i.e. `    # get max supply of this faucet. We assume it is stored at pos 3 of slot 1
`
- indicate the new stack below it, i.e., `# => [max_supply, amount, tag, RECIPIENT, ...]`

examples/dft.masm
```diff
@@ -0,0 +1,288 @@
...
+#! 0: n
+#! 1: root_of_unity
+#! 2: k
+proc.main.3
+    # Initialize memory
+    # Note: k is initialized to 0 by the VM
+    dup                                      # stack: [n, n, v0, ..., v{n-1}]
+    loc_store.0                              # stack: [n, v0, ..., v{n-1}]
+    dup                                      # stack: [n, n, v0, ..., v{n-1}]
+    mem_store.RESULT_ARR_LOC                 # stack: [n, v0, ..., v{n-1}]
```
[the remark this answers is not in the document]
plafer: I was trying to keep the semantics separate:
- `local[0]` stores the length of the array
- `RESULT_ARR_LOC` stores the result array, which happens to be encoded as `[length of array, ele_0, ele_1, ...]`

This allows us to change the encoding the of the final array without creating mysterious bugs, as well as making it more readable (i.e. we read `loc_load.0` to load the length of the array instead of `loc_load.RESULT_ARR_LOC`)

examples/dft.masm
```diff
@@ -0,0 +1,288 @@
...
+    dup                                      # stack: [n, n, v0, ..., v{n-1}]
+    exec.log2                                # stack: [log2(n), n, v0, ..., v{n-1}]
```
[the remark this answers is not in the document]
plafer: `log2` [fails](https://github.com/0xPolygonMiden/examples/blob/35e72f0beaa3b668885e2f88bec4a77dbda2eb5d/examples/dft.masm#L71) for us if the input is not a power of 2. It's also less expensive to rely on `log2` to implicitly check as opposed to e.g. performing expensive bitwise operations to check

examples/dft.masm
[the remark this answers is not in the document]
plafer: > Is this the same n as you use as the length of the input array?

No, [it's the `log`](https://github.com/0xPolygonMiden/examples/blob/35e72f0beaa3b668885e2f88bec4a77dbda2eb5d/examples/dft.masm#L242-L244) of the length of the input array

> If so, should n=32 result in const.TWO_ADIC_ROOT_OF_UNITY=7277203076849721926 ?

If the length of the array is 32, then the input to `get_root_of_unity` is 5.

> According to the comments and the Winterfell implementation, TWO_ADIC_ROOT_OF_UNITY is a primitive root of unity for 2^{32}.

👍

> But it seems to return TWO_ADIC_ROOT_OF_UNITY if log2(n) = 32, right?

Yes. So to recap, let `n` be the length of the input array, and let `w = 7277203076849721926`. 

If `n = 32`, then `get_root_of_unity` returns $w^{2^{32 - 5}} \neq w$.

If $n = 2^{32}$, then `get_root_of_unity` returns $w^{2^{32 - 32}} = w^1 = w$, as expected.

examples/dft.masm
```diff
@@ -0,0 +1,288 @@
...
+    loc_load.4                                 # load result
+end
+
+# Stores an array to `INPUT_ARR_LOC`, where an "array" is a list of elements prefixed by their length.
+# The array is stored in the same order, that is [n, v0, ..., v{n-1}].
+# The inverse operation is `proc.retrieve_array`.
+# Input: [n, v0, ..., v{n-1}, ...]
+# Output: [...]
+proc.store_array
+    # Store n
```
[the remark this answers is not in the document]
plafer: This `store_array` procedure was written to be self-contained (modulo the fact that it uses the global `INPUT_ARR_LOC` instead of receiving that address on the stack, but I got lazy); i.e. it can be used to store array of any length. So you could do (pseudocode)

```
# Storing arrays of different lengths
exec.store_array([4,1,2,3,4])
exec.store_array([3,1,2,3])
```

Basically, an array is `[length, ele_0, ..., ele_{length-1}]`, and this `mem_store` stores the `length` (which could be different from call to call).

examples/dft.masm
plafer: I believe [this comment](https://github.com/0xPolygonMiden/examples/pull/150#discussion_r1366623645) invalidates this one?

Rereading this now, I used memory to preserve on every iteration `root_of_unity`, `k` and `j`. I believe the canonical way would have been to move those values up/down with `movup`/`movdn`, as it results in fewer cycles in most cases (as one `loc_load`/`loc_store` pair costs 6-8 cycles).

examples/dft.masm
plafer: As discussed [here](https://github.com/0xPolygonMiden/examples/pull/150#discussion_r1397279064), if I were to redo, I would use stack operations instead of memory. But I don't think I should invest time in doing that at this point 🙂.

> Nice implementation and nice comments!

Thank you!

plafer:
@Dominik1999 Changed the code to follow the comment convention

examples/dft.masm
plafer: Can you check if the new version solves this problem? I think the correspondence of the stack indices of `v0` and `v{j}` is sufficient?

examples/dft.masm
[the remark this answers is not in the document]
Dominik1999: makes sense

Dominik1999 (approved):
Looks good to me, thanks!

PR #152, opened 2023-10-10
Refactor: initialising prettier and deleting the old tx proof page

Dominik1999:
Some refactoring,

I deleted the tx-proof page. We need to create a more sophisticated version for the presentation in Istanbul. 

I finally initialized `prettier`, so there are a lot of style changes.


hackaugusto (approved):
LGTM. I have not reviewed individual lines since these changes are from an automated tool. I assume the code only changed its formatting.

PR #154, opened 2023-10-23
changing the docs link to next (main is broken)

Dominik1999:
Quick change: we need to change the docs to `next`. The playground runs at `main` Miden VM 0.7 but we won't change the assembly instructions for now and `main` isn't built atm.


bobbinth (approved):
Looks good! Thank you!

PR #159, opened 2024-01-04
refactor: bumping up to Miden v0.7 to clean up a bit

Dominik1999:
Removing the transaction example and updating to Miden VM 0.7


benchmarking-cli/src/main.rs
```diff
@@ -73,12 +74,18 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let program_to_run = program.program.clone().unwrap();
 
+    let host = DefaultHost::new(MemAdviceProvider::from(inputs.advice_provider.clone()));
+    
+    let execution_options = ExecutionOptions::new(None, 64 as u32)
+    .map_err(|err| format!("{err}"))?;
+
```
phklive, line 81: I think that you could remove the `as u32` here.

benchmarking-cli/Cargo.toml
```diff
@@ -6,7 +6,8 @@ edition = "2021"
 [dependencies]
 clap = { version = "4.0.0", features = ["derive"] }
 hex = { version = "0.4.0", default-features = false }
-miden-stdlib = { version = "0.5.0", package = "miden-stdlib", default-features = false }
-miden_vm = { version = "0.6.0", package = "miden-vm", features = ["concurrent"]}
+miden-air = { version = "0.7.0", package = "miden-air", default-features = false }
```
phklive, line 9: There is a new way of adding ExecutionOptions in this version hence the need to add a dependency to miden-air now? (We didn't have it before)

playground/miden-wasm/src/transaction.rs
```diff
@@ -1,285 +0,0 @@
-use super::ExecutionProof;
```
phklive, line 1: The whole file has been deleted, the tx proving is handled directly in the lib.rs in this new version right?

phklive (approved):
LGTM! Well done, minor nits and questions.

Running correctly locally.

benchmarking-cli/Cargo.toml
Dominik1999: I couldn't get ExecutionOptions from the miden-vm. It looks like we don't export it

playground/miden-wasm/src/transaction.rs
Dominik1999: at the moment, we don't prove transactions in the playground. We soon will be, but then we use the whole client.

PR #160, opened 2024-01-07
Add binary search MASM example

polydez:
This is a Miden Assembly example code, which performs binary search for a given value in a given ascending array.  All input data should be provided in the `operands_stack` of intputs-file. If the `operands_stack` is empty, tests will run.

### DIFF
--- a/examples/bsearch.inputs
+++ b/examples/bsearch.inputs
@@ -1,0 +1,14 @@
+{
+    "note": [
+        "Provide in `operand_stack`:",
+        "1. Value to find (12)",
+        "2. Array elements (0..19)",
+        "3. Array length (20)",
+        "If n == 0 or `operand_stack` is empty, tests will run"
+    ],
+    "operand_stack": [
+        "12",
+        "19", "18", "17", "16", "15", "14", "13", "12", "11", "10", "9", "8", "7", "6", "5", "4", "3", "2", "1", "0",
+        "20"
+    ]
+}
--- a/examples/bsearch.inputs
+++ b/examples/bsearch.inputs
@@ -4,7 +4,7 @@
         "1. Value to find (12)",
         "2. Array elements (0..19)",
         "3. Array length (20)",
-        "If n == 0 or `operand_stack` is empty, tests will run"
+        "If array length is 0 or `operand_stack` is empty, tests will run"
     ],
     "operand_stack": [
         "12",
--- a/examples/bsearch.inputs
+++ b/examples/bsearch.inputs
@@ -1,11 +1,4 @@
 {
-    "note": [
-        "Provide in `operand_stack`:",
-        "1. Value to find (12)",
-        "2. Array elements (0..19)",
-        "3. Array length (20)",
-        "If array length is 0 or `operand_stack` is empty, tests will run"
-    ],
     "operand_stack": [
         "12",
         "19", "18", "17", "16", "15", "14", "13", "12", "11", "10", "9", "8", "7", "6", "5", "4", "3", "2", "1", "0",
--- a/examples/bsearch.masm
+++ b/examples/bsearch.masm
@@ -13,68 +13,95 @@ const.STARTING_MEMORY_ADDRESS=0
 #! Output: [found (true/false), addr] where `found` indicates the result (`true`/`false`),
 #!         `addr` is an address in memory of value (if it was found), or an address where value should be inserted.
 proc.binary_search
-    # Check if `start < end` then start search, otherwise prepare "not found" result and exit
-    u32assert2                              # [start, end, val, ...]
-    dup                                     # [start, start, end, val, ...]
-    dup.2                                   # [end, start, start, end, val, ...]
-    u32checked_lt                           # [start < end, start, end, val, ...]
-    if.true                                 # [start, end, val, ...]
-        push.1                              # [1, start, end, val, ...]
-    else                                    # [start, end, val, ...]
-        # Push "not found" flag and flag to skip the loop
-        push.0.0                            # [0, 0, addr, end, val, ...], addr = start
+    # check that `start` and `end` fit to `u32` type
+    u32assert2
+    # => [start, end, val, ...]
+
+    # check if `start < end`
+    dup dup.2 u32checked_lt
+    # => [start < end, start, end, val, ...]
+
+    if.true
+        # push the `1` flag to begin searching
+        push.1
+        # => [1, start, end, val, ...]
+    else
+        # push "not found" flag and flag to skip the loop
+        push.0.0
+        # => [0, 0, addr, end, val, ...], addr = start
     end
-    while.true                              # [start, end, val, ...]
-        # Calculate middle address `middle`
-        dup.1                               # [end, start, end, val, ...]
-        dup.1                               # [start, end, start, end, val, ...]
-        u32wrapping_sub                     # [count, start, end, val, ...], count = end - start
-        # calculate half of the array. If `count` is odd, the result's fraction will be discarded
-        u32checked_div.2                    # [count / 2, start, end, val, ...]
-        dup.1                               # [start, count / 2, start, end, val, ...]
-        u32wrapping_add                     # [middle, start, end, val, ...], middle = start + count / 2
 
-        dup                                 # [middle, middle, start, end, val, ...]
-        mem_load                            # [memory[middle], middle, start, end, val, ...]
+    while.true
+        # => [start, end, val, ...]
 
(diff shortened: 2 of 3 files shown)


examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+#! For input and output see documentation for `proc.binary_search_stack`.
+#!
+#! If input values are omitted or `count = 0`, tests will run.
+
+const.STARTING_MEMORY_ADDRESS=0
+
+#! Binary searches for value `val` in the memory addresses range [start..end), sorted in ascending order.
+#!
+#! Input: [start, end, val, ...]
+#! Output: [found (true/false), addr] where `found` is searching result (`true`/`false`),
```
Dominik1999, line 13: > where `found` is searching result (`true`/`false`)

nit: maybe better "where `found` indicates the result

examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+#!         `addr` is an address of value (if it was found), or an address where value should be inserted.
```
Dominik1999, line 14: > `addr` is an address of value (if it was found), or an address where value should be inserted.

nit: can you add address in memory?

examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+#! Input: [count, arr[0], arr[1], ..., arr[count - 1], val],
+#!        where:
+#!            `count` is number of elements in the array,
+#!            `arr[0]`, `arr[1]`, ..., `arr[count - 1]` are array elements,
+#!            `val` is value to search.
+#! Output: [found, index] where `found` is searching result (`true`/`false`),
+#!         index `[0, count)` of `val`, if `val` was found; otherwise index in array, where `val` should be inserted.
+proc.binary_search_stack
+    push.STARTING_MEMORY_ADDRESS            # [STARTING_MEMORY_ADDRESS, count, arr[0], arr[1], ..., arr[count - 1], val, ...]
+    exec.read_stack_to_memory               # [end, val, ...]
```
Dominik1999, line 113: should it be `write_stack_to_memort`?

examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+    drop                                    # [addr, found, val, ...]
+    swap.2                                  # [val, found, addr, ...]
+    drop                                    # [found, addr, ...]
+end
+
+#! Reads `count` values (`arr[0], arr[1], ..., arr[count - 1]`) from the stack into the memory starting from address `addr`.
+#!
+#! Input: [addr, count, arr[0], arr[1], ..., arr[count - 1], ...]
+#! Output: [end, ...] where `end` is addr next vacant memory address, thus array occupies addresses `[addr..end)` in the memory
+proc.read_stack_to_memory
```
Dominik1999, line 83: same here, should it be `write_stack_to_memory`?

examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+    u32assert2                              # [addr, count, arr[0], arr[1], ...]
+    dup.1                                   # [count, addr, c, arr[0], arr[1], ...]
```
Dominik1999, line 85: nit: `[count, addr, count, arr[0], arr[1], ...]`

examples/bsearch.masm
Dominik1999, line 83: Nice procedure to store arbitrary length

examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+proc.binary_search
+    # Check if `start < end` then start search, otherwise prepare "not found" result and exit
+    u32assert2                              # [start, end, val, ...]
+    dup                                     # [start, start, end, val, ...]
+    dup.2                                   # [end, start, start, end, val, ...]
+    u32checked_lt                           # [start < end, start, end, val, ...]
+    if.true                                 # [start, end, val, ...]
+        push.1                              # [1, start, end, val, ...]
+    else                                    # [start, end, val, ...]
+        # Push "not found" flag and flag to skip the loop
```
Dominik1999, line 24: This would be an error, right? 

You could use `assert.err=123` and break.

examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+        push.0.0                            # [0, 0, addr, end, val, ...], addr = start
+    end
+    while.true                              # [start, end, val, ...]
+        # Calculate middle address `middle`
+        dup.1                               # [end, start, end, val, ...]
+        dup.1                               # [start, end, start, end, val, ...]
+        u32wrapping_sub                     # [count, start, end, val, ...], count = end - start
+        u32checked_div.2                    # [count / 2, start, end, val, ...]
```
Dominik1999, line 32: nit: can you add in the comment what happens if `count` is odd?

examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+
+        dup                                 # [memory[middle], memory[middle], middle, start, end, val, ...]
+        dup.5                               # [val, memory[middle], memory[middle], middle, start, end, val, ...]
+        eq                                  # [memory[middle] == val, memory[middle], middle, start, end, val, ...]
+        if.true                             # [memory[middle], middle, start, end, val, ...]
+            drop                            # [middle, start, end, val, ...]
+            swap                            # [start, middle, end, val, ...]
+            drop                            # [middle, end, val, ...]
+            # Prepare "found" flag and exit the loop
+            push.1.0                        # [0, 1, addr, end, val, ...], addr = middle
```
Dominik1999, line 47: I like the usage of the flags like this

examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+    call.test_gaps_left_found
+    call.test_gaps_right_not_found
+    call.test_gaps_left_not_found
+end
+
+begin
+    dup                                     # [count, count, ...]
+    neq.0                                   # [count != 0, count, ...]
+    if.true                                 # [count, arr[0], arr[1], ..., arr[count - 1], val, ...]
+        exec.binary_search_stack            # [found, index]
```
Dominik1999, line 203: can you maybe get rid of the `found` in the final output?

You could return all `0` if you didn't find the value and the pos of the stack if you found it

examples/bsearch.masm
```diff
@@ -0,0 +1,208 @@
...
+    push.STARTING_MEMORY_ADDRESS            # [start, end, val, ...], start = STARTING_MEMORY_ADDRESS
+    exec.binary_search                      # [found, addr, ...]
+
+    swap                                    # [addr, found, ...]
+    push.STARTING_MEMORY_ADDRESS            # [STARTING_MEMORY_ADDRESS, addr, found, ...]
+    u32wrapping_sub                         # [index, found, ...], index = addr - STARTING_MEMORY_ADDRESS
+    swap                                    # [found, index, ...]
+end
+
+proc.test_empty
```
Dominik1999, line 124: This is super nice that you made tests!

Dominik1999 (commented):
Thanks Denis! This is a super cool MASM program. I left a couple of comments in line. 

One point: We try to use the convention now that we first write what the next line is about and then indicate the final stack after the operations / line with `# => [...]`

like here 

```
    # compute the pointer at which we should stop iterating
    dup.1 add
    # => [end_ptr, ptr, ...]
```

Can you change that quickly?
(I know it's not consistent yet with all the examples. I will change that.

We should combine this with the Bubble Sort of @phklive when merged.

examples/bsearch.masm
polydez: This is not an error. Binary search is used not only for searching, but also for adding new values into a sorted array. In a case when value wasn't found, the algorithm must return an index where this value should be inserted.

examples/bsearch.masm
polydez: I think, it wouldn't be a good idea to return indexes of the stack. We work with index in array, which is passed through the stack. If value is not found, `index == 0` would conflict with the index of 0th element. I tried to use special index like `0xFFFFFFFF` for such cases, but it made the code worse.

Dominik1999: agreed, thanks for the explanation

examples/bsearch.inputs
```diff
@@ -0,0 +1,14 @@
+{
+    "note": [
```
Dominik1999, line 2: Can you move those "note" comments to the beginning of the bsearch.masm file? It doesn't work on the playground (yet) having "notes" in the input

Dominik1999 (approved):
Looks good to me, thanks!

phklive:
Congrats!

PR #161, opened 2024-01-15
adding cycles and program_hash to Outputs

Dominik1999:
We add `program_hash` and `cycles` to the `Outputs` when a program runs or is proven.

We don't read the value in TypeScript-land yet, because there is a team re-writing / re-designing and they will pick it up.


hackaugusto (approved):
LGTM

PR #164, opened 2024-02-19
changes to GH actions v4 and some nits npx was complaining about

Dominik1999:
GH pages wasn't able to build the new playground due to 

- a deprecated version of GH actions 
- some linting problems

This PR addresses it and I tested the build.


playground/src/components/CodingEnvironment/MemoryInfo.tsx
```diff
@@ -21,7 +21,7 @@ const MemoryInfo = (props: MemoryInfoProps): JSX.Element => {
           <div className="flex flex-wrap overflow-auto h-40">
             {props.debugOutput &&
               formatMemory(props.debugOutput.memory).map(
-                (memoryString, index) => (
```
hackaugusto, line 24: was this a bug?

Dominik1999: `eslint` complained because `index` was not used

PR #165, opened 2024-02-19
Update Explainer.md

gubloon:
change background from white to black on the images

### DIFF
--- a/playground/src/markdown/Explainer.md
+++ b/playground/src/markdown/Explainer.md
@@ -6,13 +6,11 @@ This is a Builder's Playground for the Miden Virtual Machine. You can create any
 
 The Miden VM is a zero-knowledge virtual machine. A regular virtual machine consumes an initial state and a program and produces a final state.
 
-![image](https://github.com/0xPolygonMiden/examples/assets/7201911/e8f519de-f14d-40ea-bc0b-095d2ca92c64)
-
+<img width="700" alt="image" src="https://github.com/0xPolygonMiden/examples/assets/7201911/e8f519de-f14d-40ea-bc0b-095d2ca92c64">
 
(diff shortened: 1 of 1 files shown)


Dominik1999:
This is super nice! Can you also update the explainer in other parts? 

I think we can update the 
- picture with the blue and white buttons
- add a picture with our new way how to input

playground/src/markdown/Explainer.md
```diff
@@ -102,11 +104,21 @@ Want to know more? [Here](https://wiki.polygon.technology/docs/miden/user_docs/a
 
 ## OK, but what can I do now?
 
-![](https://i.imgur.com/Y6pkjzt.png)
+![image](https://github.com/gubloon/examples/assets/7201911/1ef88c9a-1dab-48b5-9cd6-bf58c4b7aea4)
```
Dominik1999, line 107: I think we need a different picture here now. Those buttons don't exist anymore. It looks like

Screenshot 2024-02-23 at 16 38 08

Dominik1999: And then 
- debug is only there on Desktop (not mobile)
- verify is in the proof info box (on mobile it is there instead of debug)

Dominik1999 (approved):
looks good, one small change need though

playground/src/markdown/Explainer.md
gubloon: 1. I replaced this image:
https://github.com/gubloon/examples/assets/7201911/1ef88c9a-1dab-48b5-9cd6-bf58c4b7aea4
with images for each section highlighting the relevant buttons according to the section.
2. Added descriptions in italic for both debug button and verify button as how they will appear in desktop and mobile.

Dominik1999: nice thanks, then let's merge

PR #167, opened 2024-02-23
refactor: adding a pseudo mobile version

Dominik1999:
Chaning few things, I need a good working version for Denver. 

- added a mobile version (no debugging, no instructions, easier verification, less real estate)
- all examples work now
- proving toast doesn't hang


gubloon:
For a mobile version, we should stop the instruction page appearing on the first screen. Because the developer will need the code area primarily and the input/output area secondarily but the instruction section is taking more space instead.
image

gubloon (approved):

phklive (approved):
LGTM!

Code looks sound.

PR #176, opened 2024-04-08
fixes few of the items in issue 166

gubloon:
- Replaced the instructions set with output and program info on the first load of the page
- Fixed the behaviour of the UI elements appeared with respective to clicks on the navigation bar
- Made the Output box to be displayed after clicking "Prove"


playground/src/pages/CodingEnvironment.tsx
```diff
@@ -283,7 +283,7 @@ export default function CodingEnvironment(): JSX.Element {
       if (inputObject.advice_stack) {
         setAdviceValue(formatBeautifyNumbersArray(inputObject.advice_stack));
         setIsAdviceStackLayoutVisible(true);
-      }
+      } // eslint-disable-next-line
     } catch (error: any) {
```
Dominik1999, line 287: Can you simply move line 288 up to line 287?

```suggestion
    } catch (error: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
```

This should work.

Dominik1999 (approved):
Looks good, I left one small comment

PR #182, opened 2024-04-30
updating Miden VM version in the playground

Dominik1999:
Bumping the Miden VM version to the latest. I also fixed a bug that the Output didn't change when `prove` was clicked.

Now, some examples and the std_lib fail. The example fail because most of the use some `u32unchecked` commands that do not exist anymore. 

It is unclear why `std_lib` and `advice_provider` example fail.


gubloon (approved):
Looks good to me.

PR #183, opened 2024-04-30
changing MASM u32 calls

Dominik1999:


bobbinth (approved):
Looks good! Thank you!

PR #186, opened 2024-05-30
examples: adding simple provable copliance example

Dominik1999:
We want to show this example at a conference


.github/workflows/gh-pages.yml
```diff
@@ -3,7 +3,7 @@ name: Deploy Miden Assembly Playground
 on:
   push:
     branches:
-      - dominik_forgot_some_procedures
```
hackaugusto, line 6: 😆

hackaugusto (approved):
LGTM

PR #188, opened 2024-05-31
removing the secrect input from the output

Dominik1999:
The example makes more sense, if we don't have the secret input as output


bobbinth (approved):
Looks good! Thank you!

If I understood the example correctly, the sanctioned account list is provided via the stack. I think this is fine for a simple example.

In a slightly more sophisticated example, we would provide a commitment to the sanction list via the stack, unhash the commitment in memory, and then compare the address to this list in a while loop. This could be a small "good-first-issue" for the future.

Issue #199, opened 2024-09-11
Fix the rendering erros in the Miden Assembly Instruction tab


Dominik1999:
In the Miden Playground (playground.miden.io) we have the Instructions tab. There are multiple formatting errors in the table. Our ideal state is that we show all Assembly instructions from the [Miden VM docs](https://0xpolygonmiden.github.io/miden-vm/user_docs/assembly/u32_operations.html#conversions-and-tests) in the same cool way as [EVM codes](https://www.evm.codes/).

1. The description of some instructions is not rendered correctly:
   <img width="1333" alt="Screenshot 2024-09-11 at 18 10 55" src="https://github.com/user-attachments/assets/2807bc53-6ca9-403a-a803-d88eb4e55bc7">

2. For some instructions, there should be a new line if two different ways are described, e.g., `adv.push_mapval` and `adv.push_mapval.s` is the same instruction (must be in the same row), but one is a short cut for `adv.push_mapval s`
   <img width="406" alt="Screenshot 2024-09-11 at 18 13 57" src="https://github.com/user-attachments/assets/3732edce-c77c-4c81-ba97-b2d415feb763">

3. I don't know if all the instructions are there. We should scan the following folders [Assembly Instructions](https://github.com/0xPolygonMiden/miden-vm/tree/main/docs/src/user_docs/assembly). Then, when we find a table with instructions we should add it to the JSON [here](https://github.com/0xPolygonMiden/examples/blob/main/playground/src/data/instructions.tsx). There is already a file that does that, but I am not sure it catches all the instructions, see [here](https://github.com/0xPolygonMiden/examples/blob/main/playground/src/data/index.js).

Dominik1999:
Was closed in #202

PR #202, opened 2024-09-23
Fix instruction tabs rendering

BrianSeong99:
This PR fixes miden playground instruction tab rendering problem, associated with issue #199 :
1. Instead of using `$gt;`, use `>` for fix the rendering problem
2. Fixed instructions new line problem: for new line in some of the assembly instructions, use `\n\n` instead of only one`\n`.
3. Checked all instructions on the [miden-vm](https://github.com/0xPolygonMiden/miden-vm/tree/main/docs/src/user_docs/assembly) repo, all instructions were included already.


Dominik1999 (approved):
Excellent work. I tested it and it works!

PR #210, opened 2025-04-14
Update to the latest VM and refactoring

Dominik1999:
This PR updates the playground to the latest VM version 0.13. 

It also refactors the backend to have one WASM-file (wasm.rs) and the pure rust backend with native functions. This refactoring allows for testing using `cargo test`.


playground/miden-wasm/src/backend.rs
```diff
@@ -0,0 +1,391 @@
...
+    assert_eq!(output.instruction, None);
+    assert_eq!(output.num_of_operations, None);
+    assert_eq!(output.operation_index, None);
+    assert_eq!(
+        output.stack,
+        vec![0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+    );
+    assert_eq!(output.memory, vec![0, 0, 1, 3, 2, 0, 3, 0]);
+
+    // breakpoint doesn't work right now in the VM
```
phklive, line 370: Should we remove the commented code?

playground/src/components/DropDown.tsx
```diff
@@ -8,15 +8,15 @@ const examples = [
...
   'catalan',
   'dft',
   'shamir_secret_share',
   'provable_compliance',
   'collatz',
   'comparison',
-  'nprime',
-  'matrix_multiplication'
+  //'nprime',
+  //'matrix_multiplication'
```
phklive, line 19: Why are we removing these examples?

examples/collatz.masm
```diff
@@ -19,4 +19,7 @@ begin
         dup push.1 neq
     end
     swap
+
+    # clean up stack to end with less than 16 elements
+    swapw dropw
```
phklive, line 24: Applies for all `swapw dropw`. 

Would it be better to use: `exec.sys::truncate_stack` ?

playground/miden-wasm/src/backend.rs
```diff
@@ -0,0 +1,391 @@
...
+
+use miden_stdlib::StdLibrary;
+use miden_vm::{DefaultHost, ProvingOptions, ExecutionProof};
+use miden_processor::ExecutionOptions;
+use alloc::string::String;
+use alloc::vec::Vec;
+
+const MAX_STACK_LENGTH: usize = 40;
+
+pub fn run_program_native(code: &str, inputs_str: &str) -> Result<Outputs, String> {
```
phklive, line 13: What does `native` mean in this context?

phklive (commented):
Had a few questions, overall PR lgtm.

playground/miden-wasm/src/backend.rs
Dominik1999: Yeah, we should probably. I was thinking about letting it in, because Paul has fixed that already in `next`, so it is in in the next release. What do you think?

examples/collatz.masm
Dominik1999: hehe yes, but then I have to import `sys`. I thought then the example is more complex when we import libraries. So for the examplemplatory benenfit a `swapw dropw` might be better?

PR #212, opened 2025-05-22
Fix name and URL

sergerad:


README.md
```diff
@@ -2,7 +2,7 @@
 
 Playground for example programs for the [Miden VM](https://github.com/0xPolygonMiden/miden-vm) in Miden assembly.
```
bobbinth, line 3: I think this link needs to be updated as well.

